PR #94, opened 2021-05-14
Document recommendations around semantic conventions

owais:
Proposal: https://github.com/signalfx/gdi-specification/discussions/93

### DIFF
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -14,7 +14,7 @@ All Splunk distributions of OpenTelemetry,
   - `telemetry.sdk.language`
 
 - SHOULD set the following resource attributes when applicable:
-  - `telemery.auto.version`
+  - `telemetry.auto.version`
 
 
 ## Splunk Resource Attributes
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -1,0 +1,27 @@
+# Semantic Conventions
+
+This document defines the OpenTelemetry and Splunk specific standard attributes for Splunk distributions of OpenTelemetry.
+
+## OpenTelemetry Resource Attributes
+
+**Description:** Required and recommended OpenTelemetry [resource semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk).
+
+All Splunk distributions of OpenTelemetry,
+
+- MUST set the following OpenTelemetry resource attributes according to the [OpenTelemetry Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk):
+  - `telemetry.sdk.name`
+  - `telemetry.sdk.version`
+  - `telemetry.sdk.language`
+
+- SHOULD set the following resource attributes when applicable:
+  - `telemery.auto.version`
+
+
+## Splunk Resource Attributes
+
+**Description:** Set of attributes used to uniquely identify a Splunk distro version in combination with OpenTelemetry's `telemetry.sdk.*` attributes.
+
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `splunk.distro.version` | string | The version number of the Splunk distribution being used. | `1.5.0` | Yes |
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ expectations. It is applicable to GDI projects.
 The following components are currently in scope:
 
 - [Configuration](specification/configuration.md)
+- [Semantic Conventions](specification/semantic_conventions.md)
 - [Repository](specification/repository.md)
 - [Versioning](specification/versioning.md)
 


specification/semantic_conventions.md
```diff
@@ -0,0 +1,27 @@
...
+
+
+## Splunk Resource Attributes
+
+**Description:** Set of attributes used to uniquely identify a Splunk distro version in combination with OpenTelemetry's `telemetry.sdk.*` attributes.
+
+
+| Attribute  | Type | Description  | Examples  | Required |
+|---|---|---|---|---|
+| `splunk.distro.version` | string | The version number of the Splunk distribution being used. | `1.5.0` | Yes |
```
pellared, line 27: Do we have a definition what a "Splunk distro" is?
For me it is some convenience library for creating the global tracer. Is my intuition correct? Maybe it would good to make it more clear.

Do we have (or want) any attributes for the instrumentation libraries so that we could identify which are used?

specification/semantic_conventions.md
```diff
@@ -0,0 +1,27 @@
...
+
+All Splunk distributions of OpenTelemetry,
+
+- MUST set the following OpenTelemetry resource attributes according to the [OpenTelemetry Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk):
+  - `telemetry.sdk.name`
+  - `telemetry.sdk.version`
+  - `telemetry.sdk.language`
+
+- SHOULD set the following resource attributes when applicable:
+  - `telemery.auto.version`
```
pellared, line 17:
```suggestion
- MUST set the following OpenTelemetry resource attributes according to the [OpenTelemetry Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.3.0/specification/resource/semantic_conventions#telemetry-sdk).
```

We can get rid of copy-paste of the specs by can just making a "pinned" hyperlink 😉

specification/semantic_conventions.md
```diff
@@ -0,0 +1,27 @@
+# Semantic Conventions
```
pellared, line 1: How about adding a hyperlink to this file in the main `README.md`?

specification/semantic_conventions.md
owais: > to make it clear by adding a definition here: https://github.com/signalfx/gdi-specification#terms

May be but this PR shouldn't care about that IMO.

> Do we have (or want) any attributes for the instrumentation libraries so that we could identify which are used?

Already exists:

https://github.com/open-telemetry/oteps/blob/main/text/0083-component.md#internal-details
https://github.com/open-telemetry/opentelemetry-specification/blob/d001d5d25ea11a28bd4c4ec2652e2db1a41fc064/specification/trace/sdk_exporters/non-otlp.md#instrumentationlibrary

specification/semantic_conventions.md
owais: What if the spec changes and add new fields? That'd change GDI spec on the fly without even a version number change :) I don't see any much harm in explicitly mentioning the required fields for Splunk GDI spec and linking to otel for details.

pellared: Take notice the `tree/v1.3.0`  in the hyperlink 😉

owais: Then it won't change and mentioning the attributes inline has no risk so I'd make it easier for the readers and not have them open another page to read 3 words :)

owais:
@signalfx/gdi-java-maintainers @signalfx/gdi-go-maintainers @signalfx/gdi-js-maintainers PTAL ^

flands (approved):
LGTM

pellared (approved):

specification/semantic_conventions.md
mateuszrzeszutek, line 17: Typo:

```suggestion
  - `telemetry.auto.version`
```

specification/semantic_conventions.md
mateuszrzeszutek, line 27: `Required = Yes`

Does that mean that we MUST add this attribute?

mateuszrzeszutek:
|In the javaagent we already use `telemetry.auto.version = splunk-0.11.0-otel-1.2.0` - this version is propagated to all othel instrumentations present in the javaagent. I fear that if we make this convention required (and I agree that it makes more sense to distinguish between instrumentation library, the agent and our distro) there's no way that we'll fix this in the javaagent for GA.

owais:
> ... here's no way that we'll fix this in the javaagent for GA.

Could you clarify this a bit more? Why can't we fix it in javaagent? Because of it being a "breaking change" or some other technical reason? I don't think this is a breaking change since semantic conventions are not considered stable yet AFAIK.


> this version is propagated to all otel instrumentations present in the javaagent
Out of curiosity, wouldn't this just be added to the resource set on the global provider just once instead of passing down to all instrumentations?

mateuszrzeszutek:
> Could you clarify this a bit more? Why can't we fix it in javaagent? Because of it being a "breaking change" or some other technical reason?

We can, but currently it's a bit of a mess in the javaagent: we basically take `Implementation-Version` from the Jar manifest and use it everywhere we can, without any distinction between instrumentation libraries, agents and what not.
So when we set our own Jar version it's then automagically passed everywhere.

mateuszrzeszutek:
> Out of curiosity, wouldn't this just be added to the resource set on the global provider just once instead of passing down to all instrumentations?

Yeah, we could just add a new Resource with `splunk.distro.version = 0.42.0` and that would work fine - but that would not change the fact that `telemetry.auto.version` and instrumentation library versions are just taken from the jar.

owais:
Thanks for the explanation. I think we can treat "fixing" `telemetry.auto.version` as a separate work item and do it independently. TBH it might not even require fixing in case of splunk-otel-java as it probably reflects reality already i.e, splunk-otel-java ships it's own instrumentation agent and we record it's version.

So IMO, we just need to add `splunk.distro.version` attribute to the global tracer provider's resource to comply with the proposed spec requirement.

mateuszrzeszutek (approved):

mateuszrzeszutek:
> I think we can treat "fixing" telemetry.auto.version as a separate work item and do it independently.

👍

MrAlias (approved):

PR #96, opened 2021-05-19
Clarify experimental requirements

flands:
- Define and use `component` consistently
- Require experimental components to be opt-in
- Define component stability

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,21 +9,22 @@ expectations. It is applicable to GDI projects.
 > handled upstream. For information on OpenTelemetry, see the [OpenTelemetry
 > Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md)
 
-The following components are currently in scope:
+The following specification sections are currently in scope:
 
 - [Configuration](specification/configuration.md)
 - [Repository](specification/repository.md)
 - [Versioning](specification/versioning.md)
 
-GDI projects MUST adopt GDI specification changes by their next minor release
+GDI projects MUST adopt GDI specification changes by their next `MINOR` release
 and within three months (whichever is sooner). The GDI specification and GDI
-projects MUST remove any deprecated specification at the next major release.
+projects MUST remove any deprecated specification at the next `MAJOR` release.
 
 The GDI specification is meant for GDI projects and MUST NOT be required
 reading by GDI project consumers.
 
 ## Terms
 
+- Component: Specific feature-set of a GDI project.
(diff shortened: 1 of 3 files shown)


iNikem (approved):

README.md
```diff
@@ -9,31 +9,32 @@ expectations. It is applicable to GDI projects.
...
 - GDI Project: A project in the `signalfx` GitHub that starts with
   `splunk-otel-\*`
 - Instrumentation Library: A way to emit telemetry data from an application.
   Refers to `splunk-otel-<language>` projects.
-- Maintainer: Someone responsible for the specification or a project
-- Specification: A set of requirements for projects
-- Project: A GitHub repository
+- Maintainer: Someone responsible for the specification or a project.
+- Specification: A set of requirements for projects.
+- Project: A GitHub repository.
```
pellared, line 37: Can we use the term `repository` instead of `project`?
Why?
1. There is something like GitHub Project which is something else than a GitHub Repository
1. A common definition for the project is "a temporary endeavor undertaken to create a unique result"

mateuszrzeszutek (approved):

README.md
flands: I think that is a reasonable ask -- though outside this PR. Let's file and issue and track separately.

README.md
```diff
@@ -9,31 +9,32 @@ expectations. It is applicable to GDI projects.
...
 
 - [Configuration](specification/configuration.md)
 - [Repository](specification/repository.md)
 - [Versioning](specification/versioning.md)
 
-GDI projects MUST adopt GDI specification changes by their next minor release
+GDI projects MUST adopt GDI specification changes by their next `MINOR` release
 and within three months (whichever is sooner). The GDI specification and GDI
-projects MUST remove any deprecated specification at the next major release.
+projects MUST remove any deprecated specification at the next `MAJOR` release.
```
owais, line 20: I'd frame this as `MUST NOT REMOVE before next MAJOR release` and allow keeping it even more multiple major releases.

owais (approved):
LGTM. Suggested a change but not a blocker for this PR.

README.md
pellared: I created quickly an issue so we won't forget about it: https://github.com/signalfx/gdi-specification/issues/100

pellared (approved):

README.md
MrAlias: I think this is a valid suggestion, but I don't think it is related to the changes being introduced here. Resolving this and tracking in #101 .

MrAlias (approved):

PR #97, opened 2021-05-19
Change RELEASING.md requirements

mateuszrzeszutek:
I'm not that convinced about `MUST NOT`, but let's start with stricter requirements and relax them if it's needed.

### DIFF
--- a/specification/repository.md
+++ b/specification/repository.md
@@ -98,6 +98,8 @@ approval is granted, GDI projects MUST NOT cut a GA release.
 - SHOULD have a `MIGRATING.md` if applicable
 - MUST have a `README.md`
   - MUST have a badge on the `README.md` with build status
+    - CI and PR builds and all tests/checks that are executed in them MUST be
+      publicly accessible by anyone.
   - MUST have a badge on the `README.md` with GDI specification version supported
   - SHOULD have a badge on the `README.md` with code coverage, if appropriate.
   - SHOULD have badges on the `README.md` for other relevant things including artifacts
@@ -105,7 +107,9 @@ approval is granted, GDI projects MUST NOT cut a GA release.
   - MUST have troubleshooting information in `README.md`
   - MUST have license information in `README.md`
 - MUST NOT have a `RELEASING.md` file or any release process documentation
-  - The release process is private and should not be exposed publicly.
+  - The release process involves signing built packages and binaries and thus
+    must be kept private and not exposed publicly.
+  - Exception: Go libraries
 - MUST add the [SECURITY.md](templates/SECURITY.md)
   - SHOULD add dependabot information to `SECURITY.md` if applicable
 - SHOULD NOT contain comprehensive application examples. Application examples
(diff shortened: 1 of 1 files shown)


specification/repository.md
```diff
@@ -104,8 +104,8 @@ approval is granted, GDI projects MUST NOT cut a GA release.
   - MUST have getting started information in `README.md`
   - MUST have troubleshooting information in `README.md`
   - MUST have license information in `README.md`
-- MUST have a `RELEASING.md` documenting the release process
-  - SHOULD be able to release by pushing a tag
+- MUST NOT have a `RELEASING.md` file or any release process documentation
```
pellared, line 107: I think that Go libraries could be an exception here.
@owais @MrAlias do you agree?

flands (commented):
Overall, looks good. Should add language that all tests/CI are public and maybe a note that release a private to support signing of packages.

mateuszrzeszutek:
> Should add language that all tests/CI are public and maybe a note that release a private to support signing of packages.

Done.

I've added a short "Exception: Go" note but I'd like to leave the actual explanation why it's okay to somebody that works with Go.

specification/repository.md
owais: I think we should not prevent from having a RELEASING.md file. We should restrict projects from documenting internal/private release processes instead. Projects can have a RELEASING.md file and document how to do it publicly when applicable (e.g, Go libs), have it link to internal docs or simply document that release jobs are private.

owais (changes requested):
I think we should not prevent from having a RELEASING.md file. We should restrict projects from documenting internal/private release processes instead. Projects can have a RELEASING.md file and document how to do it publicly when applicable (e.g, Go libs), have it link to internal docs, simply document that release jobs are private or skip adding the file altogether.

>  The release process involves signing built packages and binaries and thus must be kept private and not exposed publicly.

I think this ^ is probably enough and we don't need to ban having a RELEASING.md file specifically.

mateuszrzeszutek:
> Projects can have a RELEASING.md file and document how to do it publicly when applicable (e.g, Go libs), have it link to internal docs, simply document that release jobs are private or skip adding the file altogether.

Good point - done.

specification/repository.md
```diff
@@ -98,14 +98,19 @@ approval is granted, GDI projects MUST NOT cut a GA release.
...
   - SHOULD have a badge on the `README.md` with code coverage, if appropriate.
   - SHOULD have badges on the `README.md` for other relevant things including artifacts
   - MUST have getting started information in `README.md`
   - MUST have troubleshooting information in `README.md`
   - MUST have license information in `README.md`
-- MUST have a `RELEASING.md` documenting the release process
-  - SHOULD be able to release by pushing a tag
+- MAY have a `RELEASING.md` file that documents the release process, but this
+  file MUST NOT document private processes. For projects that use private release
+  jobs, the `RELEASING.md` file may be absent or just contain the following note:
```
MrAlias, line 111: This should be normative.

```suggestion
  jobs, the `RELEASING.md` file MAY be absent or, if included, just contain the following note:
```

MrAlias:
@pellared @flands PTAL

owais (approved):

pellared (approved):

PR #112, opened 2021-06-23
RUM libraries specification

mateuszrzeszutek:
First part of the RUM spec.
I tried adding only the most crucial things that seem to be common between the three libs that we already have (js-web, android, ios).

CC @johnbley @jkwatson @jtmalinowski (any RUM devs welcome!)


specification/configuration.md
```diff
@@ -130,6 +136,27 @@ beyond the OpenTelemetry specification exist.
...
+| `rumAccessToken` ()                    | RUM authentication token. [1]                                                                                                                                                                               |
+| `applicationName` ()                   | Instrumented application name. [1]                                                                                                                                                                          |
+| `globalAttributes` ()                  | OpenTelemetry [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) that will be added to every span produced by the RUM library. |
+
+- [1] These configuration properties are mandatory and MUST be provided by the user. If any of these is missing,
+  the RUM instrumentation library MUST fail to start.
+
+Other requirements:
+- RUM library MUST use the Zipkin span exporter by default
```
mateuszrzeszutek, line 157: I guess that we have a hard dependency on our RUM backend here; is this the desired state though? We could use otel resources if we didn't use zipkin and used otlp json instead

jkwatson: For the moment, it's the only way to get data into the RUM backend, so I think it can be a MUST for now (otherwise nothing will work)

specification/repository.md
```diff
@@ -144,3 +144,9 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
 - MUST document how to configure log correlation
 - MUST define minimum supported version of each auto-instrumentation framework
   - SHOULD define maximum supported version of each auto-instrumentation framework
+
+## Real User Monitoring Libraries
+
+- MUST document all supported configuration parameters and how to set them
+- MUST document how to configure manual instrumentation
+- MUST document supported instrumentations and telemetry produced by each of them
```
jkwatson, line 152: I wonder how important this part is, since the low-level telemetry may or may not be particularly accessible in the RUM UI. (the telemetry part, not the supported instrumentation)

specification/semantic_conventions.md
```diff
@@ -18,6 +18,7 @@ All Splunk distributions of OpenTelemetry,
 - SHOULD set the following resource attributes when applicable:
   - `telemetry.auto.version`
 
+Note: this section does not apply to Real User Monitoring libraries, as they do not use the OpenTelemetry Resource.
```
jkwatson, line 21: I think that we should be setting the `service.name` on the Resource, since that part does get mapped into the zipkin spans, and is visible in the RUM UI currently.

mateuszrzeszutek: We don't do it anywhere right now, do we?
Maybe we should replace the `applicationName` config with `serviceName` and use it as `service.name`, not `app`

specification/repository.md
```diff
@@ -144,3 +144,9 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
 - MUST document how to configure log correlation
 - MUST define minimum supported version of each auto-instrumentation framework
   - SHOULD define maximum supported version of each auto-instrumentation framework
+
+## Real User Monitoring Libraries
+
+- MUST document all supported configuration parameters and how to set them
```
jtmalinowski, line 150: Technicality, but, this is a bit awkward for otel-js-browser, because all ENV variables supported by Open Telemetry package itself can be set on `window` and they'll get picked up.

specification/configuration.md
```diff
@@ -130,6 +136,27 @@ beyond the OpenTelemetry specification exist.
...
+setting the following properties (in a language-specific way: Java may use builders, Swift can use named
+parameters with default values, JavaScript can use objects, etc.):
+
+| Configuration property (default value) | Description                                                                                                                                                                                                 |
+| -------------------------------------- | -----------                                                                                                                                                                                                 |
+| `beaconUrl` ()                         | RUM beacon URL, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. [1]                                                                                                                                   |
+| `rumAccessToken` ()                    | RUM authentication token. [1]                                                                                                                                                                               |
```
johnbley, line 149: We use the term `rumAuth` for this.

specification/semantic_conventions.md
```diff
@@ -27,3 +28,26 @@ All Splunk distributions of OpenTelemetry,
...
+## Real User Monitoring Spans and Attributes
+
+The following attributes MUST be added to all spans emitted by Real User Monitoring (RUM)
+instrumentation libraries:
+
+| Name                  | Type   | Description                                                           |
+| ----                  | ----   | -----------                                                           |
+| `app`                 | string | The value of the `applicationName` property configured in `SplunkRum` |
+| `splunk.rumSessionId` | string | The RUM session ID                                                    |
+| `splunk.rumVersion`   | string | Version of the RUM library                                            |
```
jtmalinowski, line 43: Shouldn't we deduplicate attribute keys if possible? So use `splunk.distro.version` here instead?

specification/configuration.md
```diff
@@ -130,6 +136,27 @@ beyond the OpenTelemetry specification exist.
...
+| `applicationName` ()                   | Instrumented application name. [1]                                                                                                                                                                          |
```
johnbley, line 150: We use the term `app` for this.

specification/configuration.md
```diff
@@ -130,6 +136,27 @@ beyond the OpenTelemetry specification exist.
...
+| `globalAttributes` ()                  | OpenTelemetry [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) that will be added to every span produced by the RUM library. |
+
```
johnbley, line 152: We should also enforce that an `environmnt` config option is available.

specification/configuration.md
```diff
@@ -130,6 +136,27 @@ beyond the OpenTelemetry specification exist.
...
+### Real User Monitoring Libraries
+
+Real User Monitoring (RUM) instrumentation libraries cannot use environment variables for configuration.
+Instead, they MUST expose a `SplunkRum` object/class/namespace (depending on the language used) that allows
+setting the following properties (in a language-specific way: Java may use builders, Swift can use named
+parameters with default values, JavaScript can use objects, etc.):
+
+| Configuration property (default value) | Description                                                                                                                                                                                                 |
+| -------------------------------------- | -----------                                                                                                                                                                                                 |
+| `beaconUrl` ()                         | RUM beacon URL, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. [1]                                                                                                                                   |
```
johnbley, line 148: Implementations MUST enforce by default that beaconUrls are `https` only, and reject/fail to start otherwise.  Implementations MAY offer an `allowInsecureBeacon` option (default `false`) for customers wishing to use that.

specification/semantic_conventions.md
jkwatson: https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java#L132

specification/configuration.md
jkwatson: Does this need to be done by the library, though? Android does this by default, so I don't know that we should be doing it on top of the android way it works.

specification/configuration.md
jkwatson: yes, I forgot this was something that we should add to the android library.

specification/configuration.md
jkwatson: on the programmatic configuration? not `appName` ?

specification/semantic_conventions.md
jkwatson: Since this isn't an distribution of OpenTelemetry, I'm not sure we should use the `distro` tag in here. This is a pretty targeted solution for RUM, based on otel, and the spans we generate won't be showing up in the normal APM UI, so I don't know that consistency is all that crucial.

That being said, I totally just made up this attribute name, and would be totally fine changing it to something else. Having some attribute to track the version of the library seems very important, though, for customer support.

specification/configuration.md
johnbley: Yep - https://github.com/signalfx/splunk-otel-js-web/blob/main/docs/Configuration.md#general-configuration-options - not built in iOS yet but let's be consistent.  In general I like it when the config name matches up with the attribute name.

specification/configuration.md
johnbley: The outcome should ideally be something the app developer can notice (an exception throw, error code returned, or error printout visible without additional config) the first time they integrate.  If Android is silently failing to send the wire traffic I'd bother to add the extra 3 lines of if/print/throw.

specification/configuration.md
jkwatson: hmm. can't say that I'm a fan of cryptic configuration method names.

specification/configuration.md
jkwatson: No, it's not silent. Android will crash the app if you try it. :)

specification/repository.md
mateuszrzeszutek: Hmm, do you think it makes sense not to describe span attributes in details? Would just general instrumentation info (when it is called, what spans/names it produces) be enough?

specification/semantic_conventions.md
mateuszrzeszutek: iOS uses `app`: https://github.com/signalfx/splunk-otel-ios/blob/main/SplunkRumWorkspace/SplunkRum/SplunkRum/InstrumentationUtils.swift#L71
JS-web seems to be using `app` too: https://github.com/signalfx/splunk-otel-js-web/blob/main/src/SplunkWebTracerProvider.ts#L53

mateuszrzeszutek: Hmm, so does android actually: https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/RumAttributeAppender.java#L56

If zipkin supports `service.name` then maybe we should switch to it in all RUM libs?

specification/repository.md
mateuszrzeszutek: I think that you can just link OTel docs and say "we support all of these too". That's what we do in instrumentation distros

specification/configuration.md
mateuszrzeszutek: 👍 

Should we use `deploymentEnvironment` option (and `deployment.environment` attribute) to match otel conventions?

Currently just `environment` is used: https://github.com/signalfx/splunk-otel-js-web/blob/main/src/index.ts#L265

specification/configuration.md
mateuszrzeszutek: The outcome is the same (app is failing), does it really matter if the check is built-in in the platform or if we do it ourselves?

specification/configuration.md
mateuszrzeszutek: Yes, but should we continue to use `rumAuth`? `rumAccessToken` is more similar to `SPLUNK_ACCESS_TOKEN` that is used in collector & instrumentation libs, and it is an "access token" (I mean you can find it on the "Access Tokens" page in the UI)

specification/configuration.md
mateuszrzeszutek: Connected to https://github.com/signalfx/gdi-specification/pull/112#discussion_r657296227: should we continue to use `app`/`appName`/`applicationName`? `serviceName` & `service.name` would be more consistent with otel conventions

specification/semantic_conventions.md
mateuszrzeszutek: How about `splunk.rum.version` and `splunk.rum.session_id`? I think that would be more in-line with the otel attribute naming strategy

specification/semantic_conventions.md
jkwatson: I think we need both at the moment, unfortunately.

mateuszrzeszutek: 😢 
We can make setting both `service.name` and `app` mandatory for now, as a preparation to switching to `service.name` in the future (if we ever are going to do that).

specification/repository.md
jkwatson: Is that something that even needs to be in each repository? I think I'd rather push all of that documentation to some kind of generalized RUM documentation site. At this point, I think users probably won't care very much about the details of the telemetry, compared to what they end up seeing in the RUM UI, etc.

specification/semantic_conventions.md
jkwatson: I'm all for that. Right now, the RUM instrumentation (which started in the browser) kind of has its own semantic conventions, not all that closely related to otel. IMO, we should converge on something at least otel-like as much as we can.

specification/repository.md
mateuszrzeszutek: That's a good point. Let's stick to a brief instrumentation documentation in the repo

specification/semantic_conventions.md
mateuszrzeszutek: > IMO, we should converge on something at least otel-like as much as we can.

Definitely! 💯 
I was hoping that this PR might be a start of this

jkwatson: For the moment, though, the RUM backend & frontend are dependent on "splunk.rumSessionId" being the name of the attribute, so we can't change that.

mateuszrzeszutek: 🤷

specification/configuration.md
johnbley: It is named differently because it is different in kind.  `rumAuth` is public and we don't want anybody pasting a random other access token into that space.

specification/configuration.md
```diff
@@ -130,6 +136,31 @@ beyond the OpenTelemetry specification exist.
...
+
+Real User Monitoring (RUM) instrumentation libraries cannot use environment variables for configuration.
+Instead, they MUST expose a `SplunkRum` object/class/namespace (depending on the language used) that allows
+setting the following properties (in a language-specific way: Java may use builders, Swift can use named
+parameters with default values, JavaScript can use objects, etc.):
+
+| Configuration property (default value) | Description                                                                                                                                                                                                 | Required |
+| -------------------------------------- | -----------                                                                                                                                                                                                 | -------- |
+| `beaconUrl` ()                         | RUM beacon URL, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. [1] [2]                                                                                                                               | Yes      |
```
jtmalinowski, line 148: 1) Is there a reason this doesn't default to us0 endpoint? 2) Could we have an alternative `beaconEndpoint = 'us0' | 'us1' | ...`, where if you set `beaconEndpoint` then you don't have to set `beaconUrl`? `beaconUrl` would still be available.

specification/configuration.md
```diff
@@ -130,6 +136,31 @@ beyond the OpenTelemetry specification exist.
...
+| `rumAccessToken` ()                    | RUM authentication token. [1]                                                                                                                                                                               | Yes      |
+| `applicationName` ()                   | Instrumented application name. [1]                                                                                                                                                                          | Yes      |
+| `globalAttributes` ()                  | OpenTelemetry [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) that will be added to every span produced by the RUM library. | Yes      |
+| `allowInsecureBeacon` (false)          | Allows sending data to insecure endpoints not using https. [2]                                                                                                                                              | No       |
```
jtmalinowski, line 152: Since we're codifying this as a spec now, I don't see a reason for this parameter at all in Browser RUM. This problem is already solved by browser vendors https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content . And provided that our backend rejects insecure incoming traffic (I don't mean redirecting to HTTPS, just straight rejection), this shouldn't be an issue for mobile RUMs either.

specification/configuration.md
jtmalinowski: Alright, but if something is clearly just an attribute (that has a semantic meaning in our backend, but doesn't have a semantic meaning in the instrumentation layer), then shouldn't we just say "add it to the global attributes"? It sort of applies similarly to application name.

specification/repository.md
jtmalinowski: What I mean is that I'm not sure we (Browser RUM) support all OTel config params / ENV variables right now in a sane way, ie. if there's an OTel option which messess up something in our configuration.

jtmalinowski:
cc @signalfx/gdi-js-web-approvers

specification/configuration.md
mateuszrzeszutek: Fair point -- I'll revert to `rumAuth`.

specification/repository.md
mateuszrzeszutek: Oh, I see. Then I would consider all options not described in the docs as unsupported; you could even add a warning "don't use OTel config directly, this project wraps over OTel and configuring the lower level SDK directly may break something"

specification/configuration.md
mateuszrzeszutek: It's a good idea, the collector already does something similar with realm/ingest URL. I'll add it, thanks!

specification/configuration.md
mateuszrzeszutek: It doesn't make sense for Android too, since it rejects all non-https connections by default. @johnbley is that option only for iOS?

johnbley: It's not just about ensuring that any traffic actually sent is https, it's about ensuring that developers very quickly know that they have mistyped something, rather than silently dropping data on the ground or allowing data to bleed out in plaintext across the internet only to be dropped by our ingest.  I would only believe it's not necessary in browser with demonstrable proof that browsers do the sane thing.  For android does it loudly and immediately complain or does it only complain later and only if you turn on debug logging?  For iOS this can be misconfigured, so, yes, it is necessary there.  It is certainly the case that we can simply not support the idea of `allowInsecureBeacon` on any given platform (particularly since some may make that impossible), but we **should** have active initialization-time checking that we are using https.

jkwatson: As mentioned above, Android will crash your app if you try to use an insecure endpoint without specifically adding the additional permission to your app manifest.

specification/configuration.md
mateuszrzeszutek: I think there's value in adding separate config options for the most commonly used attributes - it eases the setup.

We're already doing that in otel sdk: there's now `OTEL_SERVICE_NAME=...` option supported even though you could use `OTEL_RESOURCE_ATTRIBUTES=service.name=...` to define exactly the same thing.

> that has a semantic meaning in our backend, but doesn't have a semantic meaning in the instrumentation layer

It has meaning for RUM as a whole though

jtmalinowski (approved):
Looks good enough for a v1 to me.

specification/semantic_conventions.md
mhennoch: We(web rum) have a whole epic dedicated to matching OTEL semconv. Currently I think setting [service.* attributes on Resource](https://signalfuse.atlassian.net/browse/APMI-1723) is being worked on. @t2t2 knows more.

johnbley (approved):
Seconded; it's better than the lack of spec we have now.

jkwatson (approved):
thanks! looks like a great start!

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
   - Distribution MUST default to `otlp`
   - Distribution MUST offer `jaeger-thrift-splunk` that defaults to `http://127.0.0.1:9080/v1/trace`
 
+### Real User Monitoring Libraries
```
flands, line 139: nit: either this should be indented under Instrumentation Libraries or Instrumentation Libraries should further define what is in scope -- I assume it is applications and serverless.

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+
+Real User Monitoring (RUM) instrumentation libraries cannot use environment variables for configuration.
+Instead, they MUST expose a `SplunkRum` object/class/namespace (depending on the language used) that allows
+setting the following properties (in a language-specific way: Java may use builders, Swift can use named
+parameters with default values, JavaScript can use objects, etc.):
+
+| Configuration property (default value) | Description                                                                                                                                                                                                 | Required |
```
flands, line 146: nit: We are already in the Configuration section -- can we shorten the name to just Property?

```suggestion
| Property (default value) | Description                                                                                                                                                                                                 | Required |
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+| -------------------------------------- | -----------                                                                                                                                                                                                 | -------- |
+| `realm` ()                             | Splunk realm, e.g. `us0`, `us1`. If set, value of `beaconUrl` will be automatically computed based on this. [1] [2]                                                                                         | Yes      |
```
flands, line 148: Just to confirm, we are naming this `realm` and not something like `splunkRealm` because we are in the `SplunkRum` namespace?

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+| `beaconUrl` ()                         | RUM beacon URL, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. If both `realm` and `beaconUrl` are set, `beaconUrl` takes precedence. [1] [2]                                                        | Yes      |
```
flands, line 149: We seem to use `endpoint` in most places (e.g. `OTEL_EXPORTER_JAEGER_ENDPOINT` or see receivers/exporters in collector), can this be named `beaconEndpoint` for consistency?

```suggestion
| `beaconEndpoint` ()                         | RUM beacon endpoint, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. If both `realm` and `beaconEndpoint` are set, `beaconEndpoint` takes precedence. [1] [2]                                                        | Yes      |
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+| `rumAuth` ()                           | RUM authentication token. [1]                                                                                                                                                                               | Yes      |
```
flands, line 150: Same question here -- we use the term "access token" for ingest and in UI you can only define these under "access token". Totally agreed that we need to distinguish the RUM token, but perhaps we can make the name consistent.

```suggestion
| `rumAccessToken` ()                           | RUM access token. [1]                                                                                                                                                                               | Yes      |
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+| `applicationName` ()                   | Instrumented application name. [1]                                                                                                                                                                          | Yes      |
```
flands, line 151: Continuing on the consistency journey -- we use "service name".

```suggestion
| `serviceName` ()                   | Instrumented service name.                                                                                                                                                                          | Yes  [1]    |
```

specification/configuration.md
flands: Another option is to keep it as a developer flag -- do not add to README but to CONTRIBUTING and make it clear it is unsupported and may be removed.

specification/configuration.md
flands, line 148: I think the [1] should move next to the Yes for required

```suggestion
| `realm` ()                             | Splunk realm, e.g. `us0`, `us1`. If set, value of `beaconUrl` will be automatically computed based on this. [2]                                                                                         | Yes [1]      |
```

specification/configuration.md
flands, line 149: I would propose we mark realm as required and this as not required.

```suggestion
| `beaconUrl` ()                         | RUM beacon URL, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. If both `realm` and `beaconUrl` are set, `beaconUrl` takes precedence. [2]                                                        | No [1]      |
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+| `globalAttributes` ()                  | OpenTelemetry [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) that will be added to every span produced by the RUM library. | Yes      |
```
flands, line 152: Is this really required?

```suggestion
| `globalAttributes` ()                  | OpenTelemetry [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) that will be added to every span produced by the RUM library. | No      |
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+| `allowInsecureBeacon` (false)          | Allows sending data to insecure endpoints not using https. [2]                                                                                                                                              | No       |
+| `deploymentEnvironment` ()             | Sets the environment (`deployment.environment` span attribute) for all spans.                                                                                                                               | Yes      |
```
flands, line 154: We should not require this -- we can strongly recommend it

```suggestion
| `deploymentEnvironment` ()             | Sets the environment (`deployment.environment` span attribute) for all spans.                                                                                                                               | No      |
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+
+- [1] Application name, authentication token and either realm or the beacon URL MUST be provided by the user.
+  If any of these is missing, the RUM instrumentation library MUST fail to start.
+- [2] Implementations MUST enforce by default that `beaconUrl`s are https only, and reject/fail to start otherwise.
+  Implementations MAY offer an `allowInsecureBeacon` option (default false) for customers wishing to use that.
```
flands, line 159: suggestion?

```suggestion
  Implementations MAY offer an unsupported `allowInsecureBeacon` option (default false).
```

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+
+Other requirements:
```
flands, line 161: What about compression?

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+- RUM library MUST use the Zipkin span exporter by default
```
flands, line 162: v1 or v2? JSON or gRPC?

specification/configuration.md
```diff
@@ -130,6 +136,32 @@ beyond the OpenTelemetry specification exist.
...
+- RUM library MUST limit the number of sent spans to 100 in a 30 second window per `component` attribute value
```
flands, line 163: Is this customizable (up or down)? If yes, what happens if you exceed?

specification/repository.md
```diff
@@ -144,3 +144,9 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
...
+- MUST document how to configure manual instrumentation
+- MUST document supported instrumentations
```
flands, line 152:
```suggestion
- MUST document supported instrumentations
- MUST document supported platforms
```

specification/semantic_conventions.md
flands: > 😢
> We can make setting both `service.name` and `app` mandatory for now, as a preparation to switching to `service.name` in the future (if we ever are going to do that).

let's not do this -- user should set one thing: `service.name` -- if `app` is also required we should just set that for the user based on the specified `service.name`.

specification/semantic_conventions.md
```diff
@@ -27,3 +28,34 @@ All Splunk distributions of OpenTelemetry,
...
+
+Note: this section does not apply to Real User Monitoring libraries, as they do not use the OpenTelemetry Resource.
+
+## Real User Monitoring Spans and Attributes
+
+Real User Monitoring (RUM) libraries MUST set the `service.name` resource attribute to the value of the `applicationName`
+configuration property. This is the only resource attribute that RUM libraries are supposed to set because it's the only one
+the Zipkin exporter can understand.
+
+The following attributes MUST be added to all spans emitted by RUM libraries:
```
flands, line 40: What happens if they are not set?

specification/semantic_conventions.md
```diff
@@ -27,3 +28,34 @@ All Splunk distributions of OpenTelemetry,
...
+| `splunk.rum.version`  | string | Version of the RUM library                                                                                        |
+| `component`           | string | Name of the instrumentation that produced this span                                                               |
+
+The following attributes MAY be added to all spans emitted by RUM libraries:
+
+| Name                     | Type   | Description                                                                     |
+| ----                     | ----   | -----------                                                                     |
+| `deployment.environment` | string | Value of the `deploymentEnvironment` property configured in `SplunkRum`, if any |
+
+Mobile RUM libraries (iOS, Android) MUST add device and system information to all spans:
```
flands, line 55: We should have a must for browser info too!

flands (commented):
Nice!!

specification/configuration.md
jkwatson: `rumAccessToken` is what I had been using in the android configuration, so I in favor of that. :)

jkwatson: hrm. no, I was using `rumAuthToken`. Either way, Android would have to change (really not a big deal, though, whatever we decide on).

specification/configuration.md
jkwatson: In the first go-around, it's not configurable, but it would be easy to make it so. Spans get dropped beyond the limit.

specification/repository.md
jkwatson: what does "platforms" mean in this context?

specification/semantic_conventions.md
jkwatson: Good question. Do we have a default, or do we leave it out?

specification/configuration.md
jkwatson: This was a requirement of the MVP so yes?

specification/configuration.md
jkwatson: V2 JSON

Is there zipkin grpc?

specification/configuration.md
kubawach: That's an interesting topic to discuss. Typically "service" is used in a server context. RUM however is client-side. Calling this "service" instead of "application" (which is context-agnostic) may cause some misunderstanding (ie "is this name for a service that ingests the RUM data?").

specification/configuration.md
kubawach: +1 for this :)

kubawach: We could also add a rule that warning should be logged if there is a collision (both are set) - so that user knows which one will be used.

specification/semantic_conventions.md
mateuszrzeszutek: > let's not do this -- user should set one thing: `service.name` -- if `app` is also required we should just set that for the user based on the specified `service.name`.

No, what I meant by my previous comment is "RUM library should set two attributes" - user only has to set `applicationName`/`serviceName` property, its value just get placed in two attributes.

specification/configuration.md
mateuszrzeszutek: Good point - I'll put that under the Istrumentation Libraries section. Thanks!

specification/configuration.md
mateuszrzeszutek: Yeah, that was exactly what I thought - "Splunk" is already in `SplunkRum`, no need to repeat it.

specification/configuration.md
kubawach: +1 for this as well :)

specification/configuration.md
mateuszrzeszutek: I think I agree with that, although calling a mobile or a web app a "service" is a bit iffy; also see https://github.com/signalfx/gdi-specification/pull/112#discussion_r657314903

specification/configuration.md
mateuszrzeszutek: Hmm, "required" mean "RUM library MUST implement that", though I can see that it is a bit confusing. I'll split the table into two MUST/MAY tables

specification/configuration.md
mateuszrzeszutek: Do we need to keep it in the spec if it's just a developer flag? If it's just a dev setting just for the iOS library then we can probably remove it from the spec.

specification/configuration.md
mateuszrzeszutek: There was some discussion on this: https://github.com/signalfx/gdi-specification/pull/112#discussion_r657313646

I agree that "access token" is more consistent with UI, but distinguishing RUM tokens from other tokens also makes sense.

specification/configuration.md
mateuszrzeszutek: No longer applicable

specification/configuration.md
mateuszrzeszutek: RUM libraries are required to implement that property, yes

specification/configuration.md
mateuszrzeszutek: Honestly I don't know. Are there any requirements around compression? @johnbley @jkwatson @jtmalinowski

specification/configuration.md
mateuszrzeszutek: Added v2 JSON to the requirement

specification/semantic_conventions.md
mateuszrzeszutek: Hmm, what do you mean "they are not set"?
All those attributes are either something that a RUM library must implement (session id, version, component) or something that user must provide (app).

specification/configuration.md
```diff
@@ -56,10 +56,13 @@ in a large package, but makes it easy for users to switch settings via
 configuration. An advanced installation process can be provided where the user
 chooses what to install limiting the configuration options.
 
-## Environment variables
+## Configuration options
```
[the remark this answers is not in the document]
mateuszrzeszutek: The title/main header is already named `Configuration`, I think it's slightly better to specify that this section describes just supported config options.

specification/configuration.md
mateuszrzeszutek: > We could also add a rule that warning should be logged if there is a collision (both are set) - so that user knows which one will be used.

specification/semantic_conventions.md
mateuszrzeszutek: I've added `location.href` and `splunk.script_instance` (changed from `scriptInstance` to conform to OTel naming strategy).
@jtmalinowski please take a look at it, is there anything else we want to add here?

specification/configuration.md
```diff
@@ -130,6 +136,34 @@ beyond the OpenTelemetry specification exist.
...
+parameters with default values, JavaScript can use objects, etc.).
+
+RUM instrumentation libraries MUST support the following configuration properties:
+
+| Property (default value)               | Description                                                                                                                                                                                                 |
+| -------------------------------------- | -----------                                                                                                                                                                                                 |
+| `realm` ()                             | Splunk realm, e.g. `us0`, `us1`. If set, value of `beaconEndpoint` will be automatically computed based on this. [1] [2] [3]                                                                                |
+| `beaconEndpoint` ()                    | RUM beacon URL, ex. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. If both `realm` and `beaconEndpoint` are set, `beaconEndpoint` takes precedence. [1] [2] [3]                                          |
```
pellared, line 151: I suggest `e.g.` instead of `ex.` 

AFAIK `ex.` often means "excluding" and usually means "exercise" ;)

specification/repository.md
pellared: `os.type` ?

pellared (approved):

specification/configuration.md
flands: Naming is hard -- we technially call it application instrumentation versus serverless instrumentation today -- these refer to server-side. I am not sure if the brower/client is an app -- clearly that term is used on mobile devices.

specification/configuration.md
flands: My understanding is that we do not compress today -- I am curious why and the cost perf/implications (which I assume are all server-side).

specification/repository.md
flands: Yeah, perhaps wrong term -- what versions of which browsers? what version of android/iOS -- any limitations from a device perspective?

specification/semantic_conventions.md
flands: So there are two personas here 1) the library developer -- this is a must, but I assume is not enforced 2) the software developer using this instrumentation -- what if they do not set the required attributes?

specification/repository.md
jkwatson: Got it. Yep, documenting, for example, that android RUM is only supported on API level 21 and up, and requires core library desugaring for API levels 21-24 is definitely very very important.

specification/configuration.md
jkwatson: There would be additional CPU usage on the client to do compression as well. For Android, we are just doing whatever the zipkin exporter does out of the box, which is whatever the zipkin client does out of the box.

specification/configuration.md
jkwatson: Like @flands said, the user should only have to set one thing (IMO the application name, whatever we call that configuration item), and the library should do the needful under the hood and set both `app` on spans, and `service.name` on the Resource to that value.

specification/semantic_conventions.md
mateuszrzeszutek: > 1) the library developer -- this is a must, but I assume is not enforced

Well, it is enforced by being a requirement in the GDI spec.

> 2) the software developer using this instrumentation -- what if they do not set the required attributes?

They do not have to set anything in the first place - those attributes must be added by us, the library developer, the user doesn't have to do anything to have them.

mateuszrzeszutek:
Changed the following configuration option names to provoke more discussions on these two, since there seem to be a few different opinions on them:
- `rumAuth` -> `rumAccessToken`
- `applicationName` -> `serviceName`

I think I like `rumAccessToken` more because it's more consistent with the UI and "normal" instrumentation libraries.

specification/configuration.md
```diff
@@ -130,6 +136,34 @@ beyond the OpenTelemetry specification exist.
...
+| `rumAccessToken` ()                    | RUM authentication token. [1]                                                                                                                                                                               |
+| `serviceName` ()                       | Instrumented application name. [1]                                                                                                                                                                          |
+| `globalAttributes` ()                  | OpenTelemetry [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) that will be added to every span produced by the RUM library. |
+| `deploymentEnvironment` ()             | Sets the environment (`deployment.environment` span attribute) for all spans.                                                                                                                               |
```
jkwatson, line 155: Is there a default value for this? Do we leave it out if the user doesn't provide it?

specification/semantic_conventions.md
jtmalinowski: These 2 are all we want to add for now.

specification/semantic_conventions.md
```diff
@@ -27,3 +28,41 @@ All Splunk distributions of OpenTelemetry,
...
+| `device.model.name` | string | Name of the device |
+| `os.type`           | string | `iOS` or `Android` |
+| `os.version`        | string | Version of the OS  |
+
+Web RUM library MUST add the following attributes to all spans:
+
+| Name                     | Type   | Description                                                                                                                    |
+| ----                     | ----   | -----------                                                                                                                    |
+| `location.href`          | string | Value of [`location.href`](https://developer.mozilla.org/en-US/docs/Web/API/Location/href) at the moment of creating the span. |
+| `splunk.script_instance` | string | A 64bit identifier, unique to every instance of the `SplunkRum` script.                                                        |
```
jtmalinowski, line 68: Technically it's:
> every instance of `SplunkRum` library.
Because if you add the script twice, the library is still going to run once.

jtmalinowski (approved):

specification/configuration.md
mateuszrzeszutek: We leave it out - it's described in the semantic conventions doc

jkwatson:
Is there anything left to debate in here? I'd sure like to get this landed and get the android instrumentation to match this spec before the end of the PI.

pellared:
@flands PTAL and merge if you find it good enough.

kubawach:
Good for me, let's merge.

kubawach (approved):

specification/configuration.md
```diff
@@ -130,6 +136,34 @@ beyond the OpenTelemetry specification exist.
...
+| `realm` ()                             | Splunk realm, e.g. `us0`, `us1`. If set, value of `beaconEndpoint` will be automatically computed based on this. [1] [2] [3]                                                                                |
+| `beaconEndpoint` ()                    | RUM beacon URL, e.g. `https://rum-ingest.<realm>.signalfx.com/v1/rum`. If both `realm` and `beaconEndpoint` are set, `beaconEndpoint` takes precedence. [1] [2] [3]                                         |
+| `rumAccessToken` ()                    | RUM authentication token. [1]                                                                                                                                                                               |
+| `serviceName` ()                       | Instrumented application name. [1]                                                                                                                                                                          |
```
jkwatson, line 153: I don't think this is a good configuration value. I think it will be very confusing for mobile developers, since the app isn't a "service", and they will probably think this is referring to some backend service that the app might be using. I think we should use `applicationName` or `appName` or `app` instead.

jkwatson: (sorry for the late comment..I was going through to implement what was in this PR and ran into this, which had changed since I last looked in detail)

mateuszrzeszutek: I agree with you and I think that `applicationName` is a much better name - `serviceName` sounds really confusing in the context of RUM.
@flands I think you were the only person that preferred `serviceName`. Can we change that to a less confusing `applicationName`?

MrAlias: I'm guessing the ask for this to be `serviceName` was from a desire to be consistent with other parts of the specification. However, given this part of the specification in only for RUM libraries I think it is prudent to ask these kinds of questions and make sure we provide them with names that are meaningful and allow the least amount of cognitive load when they start using our services.

@flands any opposition?

specification/semantic_conventions.md
```diff
@@ -27,3 +28,41 @@ All Splunk distributions of OpenTelemetry,
 | Attribute  | Type | Description  | Examples  | Required |
 |---|---|---|---|---|
 | `splunk.distro.version` | string | The version number of the Splunk distribution being used. | `1.5.0` | Yes |
+
+Note: this section does not apply to Real User Monitoring libraries, as they do not use the OpenTelemetry Resource.
+
+## Real User Monitoring Spans and Attributes
+
+Real User Monitoring (RUM) libraries MUST set the `service.name` resource attribute to the value of the `applicationName`
```
MrAlias, line 36: We need to revert back from `serviceName` to `applicationName` or update this.

specification/semantic_conventions.md
```diff
@@ -27,3 +28,41 @@ All Splunk distributions of OpenTelemetry,
...
+configuration property. This is the only resource attribute that RUM libraries are supposed to set because it's the only one
+the Zipkin exporter can understand.
+
+The following attributes MUST be added to all spans produced by RUM libraries:
+
+| Name                  | Type   | Description                                                                                                       |
+| ----                  | ----   | -----------                                                                                                       |
+| `app`                 | string | The value of the `applicationName` property configured in `SplunkRum`; same as `service.name` resource attribute. |
+| `splunk.rumSessionId` | string | The RUM session ID                                                                                                |
```
MrAlias, line 45: It seems a bit odd this is camelCase and the version is period delimited? Should this be:

```suggestion
| `splunk.rum.sessionId` | string | The RUM session ID                                                                                                |
```

mateuszrzeszutek: Ideally it'd be `splunk.rum.session_id` (to match otel naming conventions), but it seems that this name is already hard-coded in the RUM backend and we can't really touch it.

specification/semantic_conventions.md
mateuszrzeszutek: Changed to `applicationName` everywhere, thanks!

specification/configuration.md
jkwatson: Did we really decide to stick with `serviceName`? I think this is a very bad idea, as stated above.

jkwatson: ah never mind. Looks like I was looking at at old version. :disappear:

PR #121, opened 2021-08-04
Add support for the OTEL_SERVICE_NAME env var

### LINKED ISSUE #78: Ease process for setting `service.name`
As pointed out in #25, the `service.name` resource attribute is very important for APM. This attribute is what lets APM identify individual services. Today this defaults to `unnamed-<lang>-service` for Splunk libraries. For example, if a user forgets to specify this attribute for a Node service, it'll show up as `unnamed-node-service` in APM. If a user forgets to specify it for multiple node service, then APM will likely treat all of those services as a single service.

Today the only way to specify a value for this attribute is to use the `OTEL_RESOURCE_ATTRIBUTES` environment varariable. For example. `OTEL_RESOURCE_ATTRIBUTES=service.name=svc-payments`. This can be confusing and not obvious for new users. This has been identified in the OTel community and there is an [open issue](https://github.com/open-telemetry/opentelemetry-specification/issues/709) looking to address this.

Ideally the OTel community provides a way to easily set the service name and handles any priority resolution with `OTEL_RESOURCE_ATTRIBUTES` if they add support for an additional environment variable. If this does not happen, and we discover this is a major pain point for customers, we need to add our own environment variable to support this.

### Prior Art

* Old SignalFx libraries had a dedicated config var for service name.
* Lightstep's new Otel distros have dedicated config vars for service name.
* DataDog's new Otel libraries have a dedicated config var for service name. 
* AWS distros don't do anything special and expect the user to configure the attribute via `OTEL_RESOURCE_ATTRS`. 
* Honeycomb uses the `SERVICE_NAME` environment variable that they defined to set this.

MrAlias:
Resolve #78

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -110,17 +110,25 @@ beyond the OpenTelemetry specification exist.
 
 #### [OpenTelemetry Environment Variable](https://github.com/open-telemetry/opentelemetry-specification/blob/f228a83e652e5cd3ba96b9f780b704ee7a7daa4c/specification/sdk-environment-variables.md)
 
-- `OTEL_RESOURCE_ATTRIBUTES`
-  - User MUST define [`service.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service)
-    resource attribute.
-  - Distribution MUST log a warning when the `service.name` resource attribute is not set. The
-    warning message MUST clearly describe how to set the attribute or link to relevant documentation.
-    E.g.
+- `OTEL_SERVICE_NAME`
+  - Users MUST define a name for the service they are instrumenting. The
+    service name can either be defined using the `OTEL_SERVICE_NAME` or
+    `OTEL_RESOURCE_ATTRIBUTES` environment variable. If the user fails to
+    define a service name with these environment variables the distribution
+    MUST log a warning. The warning message MUST clearly describe how to set
+    the attribute or link to relevant documentation. E.g.
     ```
     service.name attribute is not set, your service is unnamed and will be difficult to identify.
-    set your service name using the OTEL_RESOURCE_ATTRIBUTES environment variable.
-    E.g. `OTEL_RESOURCE_ATTRIBUTES="service.name=<YOUR_SERVICE_NAME_HERE>"`
+    set your service name using the OTEL_SERVICE_NAME environment variable.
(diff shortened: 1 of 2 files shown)


iNikem (approved):

mateuszrzeszutek (approved):

specification/configuration.md
```diff
@@ -110,17 +110,25 @@ beyond the OpenTelemetry specification exist.
...
-  - User MUST define [`service.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service)
-    resource attribute.
-  - Distribution MUST log a warning when the `service.name` resource attribute is not set. The
-    warning message MUST clearly describe how to set the attribute or link to relevant documentation.
-    E.g.
+- `OTEL_SERVICE_NAME`
+  - Users MUST define a name for the service they are instrumenting. The
+    service name can either be defined using the `OTEL_SERVICE_NAME` or
+    `OTEL_RESOURCE_ATTRIBUTES` environment variable. If the user fails to
+    define a service name with these environment variables the distribution
```
pellared, line 117: And what if the user uses API to set the service name instead of env var?

specification/configuration.md
```diff
@@ -110,17 +110,29 @@ beyond the OpenTelemetry specification exist.
...
     service.name attribute is not set, your service is unnamed and will be difficult to identify.
-    set your service name using the OTEL_RESOURCE_ATTRIBUTES environment variable.
-    E.g. `OTEL_RESOURCE_ATTRIBUTES="service.name=<YOUR_SERVICE_NAME_HERE>"`
+    set your service name using the OTEL_SERVICE_NAME environment variable.
+    E.g. `OTEL_SERVICE_NAME="<YOUR_SERVICE_NAME_HERE>"`
     ```
+  - The service name value MUST be used for the
+    [`service.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service)
+    resource attribute.
+  - If a service name is defined with an environment variable and in the
```
pellared, line 128: I think that at least in Otel Go and .NET SDK it the other way

MrAlias: Right. I pulled out the overlap with the OTel spec and will rely on the specification for assignment and precedence.

owais (approved):

pellared (approved):

PR #124, opened 2021-08-26
Add OTEL_ATTRIBUTE_VALUE_LENTH_LIMIT to spec

owais:
Also update limit unset value to `""`

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -129,7 +129,8 @@ beyond the OpenTelemetry specification exist.
   - Distribution MUST support and document how to switch to `b3multi`
 - Span Collection Limits
   - Distribution MUST default to `1000` for `OTEL_SPAN_LINK_COUNT_LIMIT` (not OpenTelemetry default)
-  - Distribution MUST be `unset` (unlimited) for all others (not OpenTelemetry default)
+  - Distribution MUST default to `1200` for `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` (not OpenTelemetry default)
+  - Distribution MUST be unset `""` (unlimited) for all others (not OpenTelemetry default)
 - Zipkin exporter
   - Distribution MUST NOT list Zipkin exporter as supported (not supported by Smart Agent)
 - `OTEL_TRACES_EXPORTER`


specification/configuration.md
```diff
@@ -129,7 +129,8 @@ beyond the OpenTelemetry specification exist.
   - Distribution MUST support and document how to switch to `b3multi`
 - Span Collection Limits
   - Distribution MUST default to `1000` for `OTEL_SPAN_LINK_COUNT_LIMIT` (not OpenTelemetry default)
-  - Distribution MUST be `unset` (unlimited) for all others (not OpenTelemetry default)
+  - Distribution MUST default to `1200` for `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` (not OpenTelemetry default)
+  - Distribution MUST be unset `""` (unlimited) for all others (not OpenTelemetry default)
```
pellared, line 133: 1. This is about the env vars? Correct? Not everyone is setting these values via env vars.
2. `unset` usually means that you remove the env var. Here we are setting an empty value.
3. What is the reason to NOT use OTel defaults?

owais: > This is about the env vars? Correct? Not everyone is setting these values via env vars.

I think this just means setting defaults for env var. Users can override the value either via env var or via code. Both cases should work. 

> unset usually means that you remove the env var. Here we are setting an empty value.

According to Otel spec, unset (unlimited) means setting the env var to an empty value (empty string). Not setting the env var means it'll take the default value.

> What is the reason to NOT use the OTel defaults?

We think these defaults make more sense for our customers and our product/backend. If you want specific details, I'd suggest to discuss on an internal forum :)

pellared: I see. WDYT about explicitly listing the env vars which should be unset (assuming that they are supported)?

owais: I'm fine either way but it is not strictly related to this PR so I'd open another issue/PR for it :)

pellared (approved):

specification/configuration.md
```diff
@@ -129,7 +129,8 @@ beyond the OpenTelemetry specification exist.
   - Distribution MUST support and document how to switch to `b3multi`
 - Span Collection Limits
   - Distribution MUST default to `1000` for `OTEL_SPAN_LINK_COUNT_LIMIT` (not OpenTelemetry default)
-  - Distribution MUST be `unset` (unlimited) for all others (not OpenTelemetry default)
+  - Distribution MUST default to `1200` for `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` (not OpenTelemetry default)
```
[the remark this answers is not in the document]
iNikem: I vaguely remember that "unlimited" was explicitly requested for splunk-otel-java to "preserve SFx behaviour"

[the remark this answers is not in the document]
owais: Unlimited number of attributes, events, logs, etc but not unlimited length of a single attribute. 

Python and Node use 1200 but Java uses 12000. I don't know which is the intended one. 

https://github.com/signalfx/signalfx-nodejs-tracing/#configuration-values
https://github.com/signalfx/signalfx-python-tracing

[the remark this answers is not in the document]
iNikem: > Unlimited number of attributes, events, logs, etc but not unlimited length of a single attribute.

A, makes sense.

pellared:
@owais Commit GPG sign is missing 😉 
@flands @MrAlias PTAL

MrAlias (approved):

PR #134, opened 2021-10-05
Initial profiling spec for GDI.

breedx-splk:
Hello GDI. Building on some of the work in `splunk-otel-java`, I'd like to start fleshing out some ideas for other agents to consider. This is the first stab at getting a profiling spec for Splunk GDI agents in place.

Feedback/critique is appreciated. 

Specifically, I think the call stack format might be too rooted in Java specifics and we might be better off with an implementation-specific call stack format an a hint to the backend via attribute. You'll probably pick up on this, but I could also use some help trying to decide when to classify things as "otel" vs. "splunk".

Thanks for looking.

### DIFF
--- a/specification/profiling.md
+++ b/specification/profiling.md
@@ -1,0 +1,169 @@
+# Profiling Conventions
+
+**Status**: [experimental](../README.md#versioning-and-status-of-the-specification)
+
+This document describes the specific behavior and conventions for Splunk
+agents that contain profiling features.
+
+- [Profiling Conventions](#profiling-conventions)
+    * [Activation](#activation)
+    * [Call Stack Sampling](#call-stack-sampling)
+        + [Call Stack Filtering](#call-stack-filtering)
+        + [Call Stack Span Association](#call-stack-span-association)
+        + [Call Stack Ingest](#call-stack-ingest)
+    * [ResourceLogs](#resourcelogs)
+        - [InstrumentationLibraryLogs](#instrumentationlibrarylogs)
+        - [LogRecord Common Attributes](#logrecord-common-attributes)
+        - [LogRecord Fields](#logrecord-fields)
+            * [Call Stack Format](#call-stack-format)
+- [References](#references)
+
+## Activation
+
+Agents may choose to activate or deactivate their profiler by default.
+A configuration setting MUST be provided to allow overriding the default: 
+
+* `SPLUNK_PROFILER_ENABLED` environment variable
+
+The default setting for `SPLUNK_PROFILER_ENABLED` MUST be clearly documented. 
+The agent SHOULD NOT allow changing the setting at runtime, and the initial 
+setting SHOULD be used for the entire lifespan of the application run.
+
+An agent whose profiling capability is deactivated MUST NOT introduce
+additional profiling-based overhead. It also MUST NOT ingest profiling-based
+data.
+
+## Call Stack Sampling
+
+An agent that has profiling capabilities MUST be able to sample call stacks
+at a fixed interval. This default interval MUST default to 1 second.
+
+If a language agent runtime supports threading, stacks MUST be sampled across
+all process threads. The samples for all threads SHOULD be taken instantaneously
+and, in the event that this is not feasible, MUST be taken as close together
+as possible. If the samples are taken consecutively, then the profiler MUST 
+use a consistent ordering strategy (such as thread name or ID) when sampling all
+threads.
+
+### Call Stack Filtering
+
+Various runtimes MAY contain internal and other threads that are undesirable
+to include for profiling. This could include threads that are internal to 
+runtime behavior or agent internal workings. The choice of which threads are 
+undesirable is implementation specific and not defined.
+
+The profiler SHOULD not collect call stacks from undesirable threads.
+If this is not possible, the profiler SHOULD filter these out afterward 
+and omit these call stacks from ingest. 
+
+### Call Stack Span Association
+
+When a call stack is sampled during the execution of a span scope,
+the profiler MUST be able to associate the call stack to the span.
+This association SHOULD happen as close to the sampling point as feasible,
+but MAY occur later in a processing pipeline.
+
+After the association has been made, the `TraceId` and `SpanId` fields of the 
+LogRecord MUST be populated (see below).
+
+### Call Stack Ingest
+
+Call stacks MUST be ingested as [OpenTelemetry Logs](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs).
+The logs containing profiling data MUST be sent via OTLP/gRPC. 
+
+By default, the agent MUST send logs to the value in the `OTEL_EXPORTER_OTLP_ENDPOINT` environment
+variable. If this variable is not set, then the agent must default back to the gRPC/OTLP
+default (https://localhost:4317).
+
+The agent must allow the destination for profiling logs to be overridden with 
+the environment variable SPLUNK_PROFILER_LOGS_ENDPOINT.
+
+Agents SHOULD reuse persistent gRPC/OTLP connections from other components (traces, metrics).
+
+## ResourceLogs
+
+ResourceLogs is the upstream protobuf data type. It MUST be populated with the attributes
+from the OpenTelemetry resource.
+
+#### InstrumentationLibraryLogs
+
+Each `ResourceLogs` instance has an instance of `InstrumenationLibraryLogs`. 
+For each `InstrumentationLibraryLogs` instance:
+
+* `name` - MUST be set to `otel.profiling`
+* `version` - MUST be set to `0.1.0` 
+
+#### LogRecord Common Attributes
+
+Inside each `InstrumentationLibraryLogs` instance is a list of `LogRecord` instances. For each 
+`LogRecord` instance:
+
+* `com.splunk.sourcetype` MUST be set to the value `otel.profiling`
+* `source.event.period` MUST contain the sampling period for call stack sampler
+* `source.event.name` OPTIONALLY can contain the name of the event that triggered the sampling
+
+#### LogRecord Fields
+
+* [Name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-name)
+  MUST be set to the constant value `otel.profiling`.
+* [Time](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp)
+  MUST be set to the time that the call stack was sampled.
+* [TraceId](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-traceid)
+  MUST be populated when a call stack has been sampled within a span scope.
+* [SpanId](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-spanid)
+  MUST be populated when a call stack has been sampled within a span scope.
+* [Body](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body)
+  MUST be populated with the line-terminated call stack (see below)
+
+##### Call Stack Format
+
+The call stack is a series of lines separated by newlines (`\n`).
+The first line of the call stack is a REQUIRED thread metadata line.
+An example metadata is shown here:
+```
+"pool-1-thread-30" #49 prio=5 os_prio=31 cpu=0.12ms elapsed=8.50s tid=0x00007fde4d00c000 nid=0xba03 waiting on condition  [0x000070000ba40000]
+```
+The metadata line consists of several space separated fields.
+* thread name, surrounded with double quotes `"`
+* `#<n>` index of this thread
+* `prio=<n>` where `<n>` is the thread priority. 0 if not available.
+* `os_prio=<n>` where `<n>` is the operating system thread priority. 0 if not available.
+* `cpu=<n><u>` where `<n>` is the amount of cpu time consumed by the thread and `<u>` is a time unit abbreviation. 0 if not available.
+* `elapsed=<n><u>` where `<n>` is the wall time that thread has been alive and `<u>` is a time unit abbreviation. 0 if not available.
+* `tid=0x<n>` where `<n>` is the lowercase hexadecimal representation of the thread id. 0 if not available.
+* `nid=0x<n>` where `<n>` is the lowercase hexadecimal representation of the native thread id. 0 if not available.
+* The remainder of the line is a free-form text field that SHOULD indicate the status of the thread at the time the stack was captured (such as "running" or "waiting" )
+
+Following the metadata line is a line containing the implementation specific thread state.
+For example, java may provide:
+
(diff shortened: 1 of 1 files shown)


specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+
+ResourceLogs is the upstream protobuf data type. It MUST be populated with the attributes
+from the OpenTelemetry resource.
+
+#### InstrumentationLibraryLogs
+
+Each `ResourceLogs` instance has an instance of `InstrumenationLibraryLogs`. 
+For each `InstrumentationLibraryLogs` instance:
+
+* `name` - MUST be set to `otel.profiling`
```
flands, line 93: Any concerns when otel has a spec here?

flands (commented):
Looks good! Are more configuration parameters needed like enabled CPU vs. memory profiling?

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Activation
+
+Agents may choose to activate or deactivate their profiler by default.
+A configuration setting MUST be provided to allow overriding the default: 
+
+* `SPLUNK_PROFILER_ENABLED` environment variable
+
+The default setting for `SPLUNK_PROFILER_ENABLED` MUST be clearly documented. 
+The agent SHOULD NOT allow changing the setting at runtime, and the initial
```
pjanotti, line 29: It simpler to say only the second part: that the agent will read the env var at initialization and will ignore any changes after the initialization. Notice that only the application itself could make "observable" changes to the env vars read by the agent (I assume that the agent is in the same process as the app).

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+provided first and the bottom of the stack provided last. The format of each stack trace
+line is:
+
+```
+        at <class>.<method>(<source-module>:<line>)
+```
+
+* OPTIONAL leading whitespace
+* OPTIONAL "at "
+* `<class>` - REQUIRED. Fully scoped name of the class or module. If the function being executed is globally scoped this MUST be
```
pjanotti, line 156: Not all languages are object oriented :) although, typically there are equivalent constructs (packages, modules, etc). Related to the next command: it should be the function/method signature according to the language.

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+  set to `global`. If class is unknown, the value MUST be set to `unknown`.
+* `<method>` - Name of the method or function being invoked
```
pjanotti, line 158: What about "`method signature` according to the stack frame language/symbols"? A typical case will be for languages that support method override via the method parameters.

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+* literal `(`
+* `<source-module>` - REQUIRED. The name of the source code file (including filename extension). If the function being
+  executed is OS native, an implementation-specific indication SHOULD be used. If the source module
+  cannot be determined (eg. when symbols have been excluded), the value MUST be `unknown`
+* literal `)`
+* literal `:` - REQUIRED if `<line>` is known, OPTIONAL otherwise.
+* `<line>` - The line of source code that corresponds
```
pjanotti, line 165: Certain symbol files can produce range of lines and columns it will be good if the standard is something that covers these cases, something like `<line-start>[:<line-end> [col-start:col-end]]` perhaps

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+as possible. If the samples are taken consecutively, then the profiler MUST 
+use a consistent ordering strategy (such as thread name or ID) when sampling all
+threads.
+
+### Call Stack Filtering
+
+Various runtimes MAY contain internal and other threads that are undesirable
+to include for profiling. This could include threads that are internal to 
+runtime behavior or agent internal workings. The choice of which threads are 
+undesirable is implementation specific and not defined.
```
pellared, line 53: I am not sure here if it is a good recommendation to exclude any threads. Especiially I am concerned about:

> This could include threads that are internal to runtime behavior or agent internal workings. 

E.g. in .NET, the profiles of GC threads are extremely valuable to detect problems with memory allocation. Without the "internal working" profiles we may have trouble finding issues connected with the internals.

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
+# Profiling Conventions
+
+**Status**: [experimental](../README.md#versioning-and-status-of-the-specification)
+
+This document describes the specific behavior and conventions for Splunk
+agents that contain profiling features.
```
pellared, line 6: Should we not replace `agent` with `instrumentation library` here and in other places

See: https://github.com/signalfx/gdi-specification#terms and the rest of the spec e.g. https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+the environment variable SPLUNK_PROFILER_LOGS_ENDPOINT.
+
+Agents SHOULD reuse persistent gRPC/OTLP connections from other components (traces, metrics).
+
+## ResourceLogs
+
+ResourceLogs is the upstream protobuf data type. It MUST be populated with the attributes
+from the OpenTelemetry resource.
+
+#### InstrumentationLibraryLogs
```
pellared, line 88: nit: why not `###` ? applies to next headers as well

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+A configuration setting MUST be provided to allow overriding the default: 
+
+* `SPLUNK_PROFILER_ENABLED` environment variable
+
+The default setting for `SPLUNK_PROFILER_ENABLED` MUST be clearly documented. 
+The agent SHOULD NOT allow changing the setting at runtime, and the initial 
+setting SHOULD be used for the entire lifespan of the application run.
+
+An agent whose profiling capability is deactivated MUST NOT introduce
+additional profiling-based overhead. It also MUST NOT ingest profiling-based
```
vovencij, line 33: `ingest`? Maybe `emit`?

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+LogRecord MUST be populated (see below).
+
+### Call Stack Ingest
+
+Call stacks MUST be ingested as [OpenTelemetry Logs](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs).
+The logs containing profiling data MUST be sent via OTLP/gRPC. 
+
+By default, the agent MUST send logs to the value in the `OTEL_EXPORTER_OTLP_ENDPOINT` environment
+variable. If this variable is not set, then the agent must default back to the gRPC/OTLP
+default (https://localhost:4317).
```
vovencij, line 76:
```suggestion
default (`https://localhost:4317`).
```

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+* [Time](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp)
+  MUST be set to the time that the call stack was sampled.
+* [TraceId](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-traceid)
+  MUST be populated when a call stack has been sampled within a span scope.
+* [SpanId](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-spanid)
+  MUST be populated when a call stack has been sampled within a span scope.
+* [Body](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body)
+  MUST be populated with the line-terminated call stack (see below)
+
+##### Call Stack Format
```
pellared, line 118: Have you been considering defining it as structured JSON or Protobuf instead? I am just thinking that maybe it would be easier to implement in other languages and it could be easier to review the format.

pellared (commented):
Nice work. I left a few comments. Feel free to resolve these which in your opinion do not make any sense 😉

specification/profiling.md
vovencij: +1

pprof uses notions of `function`, `file` and `line`. For Java `function` can be represented as a method signature. Other languages may have other ways of presenting the signature of a function.

mateuszrzeszutek: [OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#source-code-attributes) uses `namespace`, `function`, `lineno` - we could use similar/same names here for consistency.

breedx-splk:
> Looks good! Are more configuration parameters needed like enabled CPU vs. memory profiling?

Probably down the road, yes. For now, primarily targeting CPU.

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
+# Profiling Conventions
+
+**Status**: [experimental](../README.md#versioning-and-status-of-the-specification)
+
+This document describes the specific behavior and conventions for Splunk
+agents that contain profiling features.
+
+- [Profiling Conventions](#profiling-conventions)
```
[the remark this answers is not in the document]
breedx-splk: Cool, I somehow missed that so I'll remove it. 👍🏻

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+    * [ResourceLogs](#resourcelogs)
+        - [InstrumentationLibraryLogs](#instrumentationlibrarylogs)
+        - [LogRecord Common Attributes](#logrecord-common-attributes)
+        - [LogRecord Fields](#logrecord-fields)
+            * [Call Stack Format](#call-stack-format)
+- [References](#references)
+
+## Activation
+
+Agents may choose to activate or deactivate their profiler by default.
```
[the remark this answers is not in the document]
breedx-splk: Agreed. Will add this.

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+and omit these call stacks from ingest. 
+
+### Call Stack Span Association
+
+When a call stack is sampled during the execution of a span scope,
+the profiler MUST be able to associate the call stack to the span.
+This association SHOULD happen as close to the sampling point as feasible,
+but MAY occur later in a processing pipeline.
+
+After the association has been made, the `TraceId` and `SpanId` fields of the
```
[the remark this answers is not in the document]
breedx-splk: The only two fields? No, time and name fields of the LogEntry are also required (below). As far as other tracing attributes go, it is assumed that they all live within the span itself. These are only required in order to do linking and are not intended to further describe the span itself.

specification/profiling.md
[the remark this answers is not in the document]
breedx-splk: Interesting. This is the first time I've seen this come up for logs. Because upstream doesn't have a logs exporter yet (it's being developed slowly), I hoped to do the simple thing for GDI. I would expect this to change as logs become stable.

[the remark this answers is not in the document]
breedx-splk: fwiw, java doesn't yet have an OTLP HTTP exporter for logs.

specification/profiling.md
```diff
@@ -0,0 +1,169 @@
...
+The logs containing profiling data MUST be sent via OTLP/gRPC. 
+
+By default, the agent MUST send logs to the value in the `OTEL_EXPORTER_OTLP_ENDPOINT` environment
+variable. If this variable is not set, then the agent must default back to the gRPC/OTLP
+default (https://localhost:4317).
+
+The agent must allow the destination for profiling logs to be overridden with 
+the environment variable SPLUNK_PROFILER_LOGS_ENDPOINT.
+
+Agents SHOULD reuse persistent gRPC/OTLP connections from other components (traces, metrics).
```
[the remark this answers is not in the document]
breedx-splk: Java doesn't do this yet, but I want it to...because I believe that gRPC channels are backed by a connection pool and thread pool and not reusing takes up extra resources.

specification/profiling.md
breedx-splk: This is great info folks, thank you. 

I like the idea of sticking with otel terms as @mateuszrzeszutek suggested: `namespace`, `function`, `lineno`.  Will also change `source-module` to `file` as @vovencij suggested.  👍🏻 ty

specification/profiling.md
breedx-splk: Changed based on the other discussion. It's namespace + function now.

specification/profiling.md
breedx-splk: Fair enough, that's why it's a SHOULD. In the dotnet case, if those threads are crucial to memory profiling and you need to keep them, so be it. I think that would classify as a desirable thread then and still meet the spec.

specification/profiling.md
breedx-splk: Naming is hard. I like this change. 😁

breedx-splk: My tongue always favors agent because it's 6 syllables shorter!  Heh.

specification/profiling.md
breedx-splk: Oops, I think this was an artifact from when I moved things around. Thanks for pointing it out.

specification/profiling.md
breedx-splk: Yeah, I am considering that or possibly just making it implementation specific and adding a form of content-type attribute. For now, I just went with what Java has provided (and what the backend expects currently). I would like to treat this kind of change as a revision to the spec later as needed.

specification/profiling.md
breedx-splk: Yeah, perhaps...

The collector HEC exporter [is using this same term](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1de839dac2e14f98485d2f58a98b89d59bfda2a9/exporter/splunkhecexporter/client.go#L166) , so a change would need to be both coordinated between instrumentation libraries and collectors and would probably want to remain backward compatible?

specification/profiling.md
breedx-splk: Oh yeah great idea. I'll add that.

specification/profiling.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+* OPTIONAL leading whitespace
+* OPTIONAL "at "
+* `<namespace>` - REQUIRED. Fully qualified namespace of the function call. For object-oriented
+  languages this SHOULD be the module/package and class name. If the function being executed is globally 
+  scoped this MUST be set to `global`. If class is unknown, the value MUST be set to `unknown`.
+* `<function>` - Name of the function or class method being invoked
+* literal `(`
+* `<file>` - REQUIRED. The name of the source code file (including filename extension). If the function being
```
mateuszrzeszutek, line 150: Should it be `source-module` instead?
```suggestion
* `<source-module>` - REQUIRED. The name of the source code file (including filename extension). If the function being
```

specification/profiling.md
```diff
@@ -0,0 +1,162 @@
...
+  executed is OS native, an implementation-specific indication SHOULD be used. If the source module
+  cannot be determined (eg. when symbols have been excluded), the value MUST be `unknown`
+* literal `)`
```
mateuszrzeszutek, line 153: Shouldn't it be the last item in this list?

pellared (approved):

flands (approved):

PR #135, opened 2021-10-07
Serverless configuration 

ghost:
In order to unify main configuration between components (metrics extension, traces wrappers) and minimise user friction following config spec for serverless is proposed.

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -92,7 +92,7 @@ environment variables:
 | :------------------------------------:          | :--------------------------------------------------------:     |
 | `SPLUNK_ACCESS_TOKEN` ()                        | Access token added to exported data. [1]                       |
 | `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [2] |
-| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for non-OTel metrics data ingest.                     |
+| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for metrics data ingest.                     |
 
 - [1]: Not user required if another system performs the authentication. For
   example, instrumentation libraries SHOULD send data to a locally running


specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
 - RUM library MUST use the Zipkin v2 JSON span exporter by default
 - RUM library MUST limit the number of sent spans to 100 in a 30 second window per `component` attribute value
 
+#### Serverless 
+
+Current serverless offering is composed of separate metrics (SFx-based) and traces (OTel-based) solutions.
```
flands, line 168: I would propose we do not cover or talk about SFx-based stuff in GDI spec. This should be purely OTel.

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
 - RUM library MUST use the Zipkin v2 JSON span exporter by default
 - RUM library MUST limit the number of sent spans to 100 in a 30 second window per `component` attribute value
 
+#### Serverless 
+
+Current serverless offering is composed of separate metrics (SFx-based) and traces (OTel-based) solutions. 
+It is also assumed that all the components send data directly to Splunk Observability Cloud (direct ingest). Therefore a set of specific configuration properties needs to be defined for the best user experience.
+   
+All serverless solution (wrappers, extensions and other vendor-specific ones) MUST honour following environment variables:
```
flands, line 171:
```suggestion
All serverless solution (wrappers, extensions and other vendor-specific ones) MUST honour the following environment variables: 
```

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
...
+
+Current serverless offering is composed of separate metrics (SFx-based) and traces (OTel-based) solutions. 
+It is also assumed that all the components send data directly to Splunk Observability Cloud (direct ingest). Therefore a set of specific configuration properties needs to be defined for the best user experience.
+   
+All serverless solution (wrappers, extensions and other vendor-specific ones) MUST honour following environment variables: 
+
+| Name                                   | Default value                                                              | Description                                                   |
+| -------------------------------------- | -------------------                                                        | ------------------------------------------------------------- |
+| `SPLUNK_ACCESS_TOKEN`                  | None                                                                       | Access token added to exported data.                          |
+| `SPLUNK_REALM`                         | `us0`                                                                      | Realm configured for the exporter endpoint. [1]               |
```
flands, line 176: I am surprised this is not part of https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries... isn't it required to support direct ingest?

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
...
+| `SPLUNK_DEBUG`                         | false                                                                      | If set to true, additional debug information will be logged   |
```
flands, line 177: Is this required? I believe some languages do not have logs today (e.g. Go). Also, why for serverless but not instrumentation?

flands (commented):
General question here -- I am not really seeing anything specific to serverless. Can we just combine instrumentation and serverless and require they both be the same?

specification/configuration.md
kubawach: I guess that for instrumentation libraries URL is enough as it already contains the realm (URLs are realm-specific).

kubawach:
> General question here -- I am not really seeing anything specific to serverless. Can we just combine instrumentation and serverless and require they both be the same?

There are basically three reasons why I think we need this and we need this separated:
1. Serverless integrates various components, some non-OTel (metrics extension). Hence need to unify basic configuration variables - like realm / ingest endpoint URL. Without this users in most basic cases would need to configure the same option twice. Also wrappers are always a bit cumbersome to configure. The goal here is to cover base uses cases with default values so most customers can jumpstart with only two env properties set.
2. Serverless aims for direct ingest - which opposes agents' collector-first approach (different default exporter and URL).
3. Serverless needs to limit logs output due to additional costs (hence DEBUG flag).

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
...
+| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` | true                                                                       | Whether Server-Timing header is added to HTTP responses       |
+
+OpenTelemetry-based wrappers MUST honour all configuration properties as defined by the OTel spec.
+Following properties MUST have default values set, as specified:
+- `OTEL_TRACES_EXPORTER` - defaults to `otlp-http`
```
mateuszrzeszutek, line 182: There's no such exporter as `otlp-http` - neither in the spec nor in the Java SDK. Perhaps you meant setting this to `otlp` and `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` to `http/protobuf`?

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
...
+- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` - defaults to `https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp` [1]
+- `OTEL_PROPAGATORS` - defaults to `tracecontext,baggage`
+- `OTEL_SPAN_LINK_COUNT_LIMIT` - defaults to 1000
+- `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` - defaults to 12000
+- `OTEL_METRICS_EXPORTER` - unset
+
+Metrics components MUST honour following configuration values and defaults:
+- `SPLUNK_METRICS_INGEST_URL` - defaults to `https://ingest.${SPLUNK_REALM}.signalfx.com` [1]
```
mateuszrzeszutek, line 190: We're using `SPLUNK_METRICS_ENDPOINT` in the javagent - it is not specced (yet), but still we should probably be consistent about it.

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
...
+| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` | true                                                                       | Whether Server-Timing header is added to HTTP responses       |
+
+OpenTelemetry-based wrappers MUST honour all configuration properties as defined by the OTel spec.
+Following properties MUST have default values set, as specified:
+- `OTEL_TRACES_EXPORTER` - defaults to `otlp-http`
+- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` - defaults to `https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp` [1]
+- `OTEL_PROPAGATORS` - defaults to `tracecontext,baggage`
+- `OTEL_SPAN_LINK_COUNT_LIMIT` - defaults to 1000
+- `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` - defaults to 12000
+- `OTEL_METRICS_EXPORTER` - unset
```
mateuszrzeszutek, line 187: Unset or `none`? Unset basically means that otlp will be used

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
...
+| `SPLUNK_REALM`                         | `us0`                                                                      | Realm configured for the exporter endpoint. [1]               |
+| `SPLUNK_DEBUG`                         | false                                                                      | If set to true, additional debug information will be logged   |
+| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` | true                                                                       | Whether Server-Timing header is added to HTTP responses       |
+
+OpenTelemetry-based wrappers MUST honour all configuration properties as defined by the OTel spec.
+Following properties MUST have default values set, as specified:
+- `OTEL_TRACES_EXPORTER` - defaults to `otlp-http`
+- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` - defaults to `https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp` [1]
+- `OTEL_PROPAGATORS` - defaults to `tracecontext,baggage`
+- `OTEL_SPAN_LINK_COUNT_LIMIT` - defaults to 1000
```
mateuszrzeszutek, line 185: Nit: does this one even matter at all? Our backend doesn't support links 😅

specification/configuration.md
kubawach: Makes sense, will change that.

specification/configuration.md
kubawach: It's copied from instrumentations spec :D

specification/configuration.md
```diff
@@ -163,6 +163,35 @@ Other requirements:
 - RUM library MUST use the Zipkin v2 JSON span exporter by default
 - RUM library MUST limit the number of sent spans to 100 in a 30 second window per `component` attribute value
 
+#### Serverless 
+
+Current serverless offering is composed of separate metrics (non OTel) and traces (OTel) solutions.
```
pellared, line 168: I suggest replacing "OTel" with "OpenTelemetry" and "spec" with "specification".

pellared (dismissed):
only some nit comments

specification/configuration.md
mateuszrzeszutek: > I can support both propositions. Adding `otlp-http` will certainly simplify life or our customers using direct ingest (also for instrumentations).

If we want to introduce this, we'd have to do it in all instrumentation distros

mateuszrzeszutek:
1. I don't think that `realm` makes that much sense in general instrumentation libraries - they're supposed to use the collector. I'm fine with keeping serverless separate.
2. Can we create a separate issue (and possibly a PR) for that? And just leave OTLP here. It's a bigger change that involves all instrumentation libraries (well, that's my opinion) and it doesn't make sense to block this PR on it.

specification/configuration.md
mateuszrzeszutek: So, can you go back to `otlp` in this PR?

pellared:
> > 2. Can we create a separate issue (and possibly a PR) for that? And just leave OTLP here. It's a bigger change that involves all instrumentation libraries (well, that's my opinion) and it doesn't make sense to block this PR on it.
> 
> Agreed, makes sense. Now we only need to wait for @flands to approve ;)

Just please change `otlp-http` to `otlp` + `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf` 😉

flands:
> > General question here -- I am not really seeing anything specific to serverless. Can we just combine instrumentation and serverless and require they both be the same?
> 
> There are basically three reasons why I think we need this and we need this separated:
> 
> 1. Serverless integrates various components, some non-OTel (metrics extension). Hence need to unify basic configuration variables - like realm / ingest endpoint URL. Without this users in most basic cases would need to configure the same option twice. Also wrappers are always a bit cumbersome to configure. The goal here is to cover base uses cases with default values so most customers can jumpstart with only two env properties set.

Doesn't splunk-otel-java do the same thing today with Micrometer? Regardless of app or serverless I would assume the configuration options are the same.

> 2. Serverless aims for direct ingest - which opposes agents' collector-first approach (different default exporter and URL).

App instrumentation must support this too -- agreed we recommend app to Collector, but once Collector extension exists same may apply to lambda given latency concerns.

> 3. Serverless needs to limit logs output due to additional costs (hence DEBUG flag).

Same applies to app no? splunk-otel-java does not enable debug logging by default

specification/configuration.md
flands: Well if we have a wrappers section it can diverge from app configuration. We could also make `otlp-http` optional. Is there no way in spec today to configure between OTLP gRPC vs. HTTP??

mateuszrzeszutek:
> We fully support and should continue to support app instrumentation direct to ingest.

Yeah, we do that. We don't really support a `SPLUNK_REALM` setting to facilitate setting up this scenario - you have to set the proper endpoints by yourself.
Should we introduce that config option everywhere? I mean, sending to collector is the default scenario in non-serverless instrumentation libraries; we didn't really think about adding a "realm" config because it seemed not to be that needed before.

kubawach:
> > 1. Serverless integrates various components, some non-OTel (metrics extension). Hence need to unify basic configuration variables - like realm / ingest endpoint URL. Without this users in most basic cases would need to configure the same option twice. Also wrappers are always a bit cumbersome to configure. The goal here is to cover base uses cases with default values so most customers can jumpstart with only two env properties set.
> 
> Doesn't splunk-otel-java do the same thing today with Micrometer? Regardless of app or serverless I would assume the configuration options are the same.
> 
Only with regards to URL (SPLUNK_METRICS_ENDPOINT - as agreed). REALM support along with defaults and DEBUG is not.

> > 1. Serverless aims for direct ingest - which opposes agents' collector-first approach (different default exporter and URL).
> 
> App instrumentation must support this too -- agreed we recommend app to Collector, but once Collector extension exists > same may apply to lambda given latency concerns.

Sure, but we want to release way before the extension is available. So need to keep different default for serverless currently :)

> 
> > 1. Serverless needs to limit logs output due to additional costs (hence DEBUG flag).
> 
> Same applies to app no? splunk-otel-java does not enable debug logging by default
>
The issue here as is all other places - instrumentations rely on OTel spec, while the AWS layer merges Splunk metrics and OTel wrappers. So we can:
- have different variables to control DEBUG for instrumentations and metrics (increases user friction)
- make metrics extension support OTel conf variable (strange)
- introduce a new one, generic for all components (what is proposed here)

specification/configuration.md
kubawach: Well.. "spec" is not used anywhere in the document  😆

theletterf: It's fine to use "OTel" throughout the text after using "OpenTelemetry".

specification/configuration.md
```diff
@@ -163,6 +163,36 @@ Other requirements:
...
+| -------------------------------------- | -------------------                                                        | ------------------------------------------------------------- |
+| `SPLUNK_ACCESS_TOKEN`                  | None                                                                       | Access token added to exported data.                          |
+| `SPLUNK_REALM`                         | `us0`                                                                      | Realm configured for the exporter endpoint. [1]               |
+| `SPLUNK_DEBUG`                         | false                                                                      | If set to true, additional debug information will be logged   |
+| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` | true                                                                       | Whether Server-Timing header is added to HTTP responses       |
+
+OpenTelemetry-based wrappers MUST honour all configuration properties as defined by the OTel spec.
```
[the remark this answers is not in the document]
kubawach: Got me! :D

mateuszrzeszutek:
I'm not against adding `SPLUNK_REALM` to library instrumentations, this could be a worthwhile addition that'd simplify connecting directly to the ingest (and establish another common link between instrumentation and collector/RUM - since they support `SPLUNK_REALM`). I don't think it's a good idea to do that in this PR though - supporting `SPLUNK_REALM` comes with its own, separate set of questions and concerns that we have to address.

owais:
`SPLUNK_REALM` does not make a lot of sense for non-RUM and non-ServerLess libraries as they work today as

- They default exporting to local collector which does not use SPLUNK_REALM
- The way to configure them to send to Splunk ingest is to set export endpoint to the full Splunk ingest endpoint which includes realm.

What would adding support for SPLUNK_REALM mean exactly? 

We could make the default exporter and export URL conditional and automatically set it to ingest URL derived from SPLUNK_REALM when the env var is set but is that so much better than just setting the explicit URL? I feel we'll end up adding unnecessary complexity both for us and end-users.

I'd suggest to keep our core otel distros simple and add SPLUNK_REALM as a concept in higher level libraries that wrap our distros such as serverless libs, zero-config agents, otel collector, helm chart, operator etc.

specification/configuration.md
```diff
@@ -165,34 +165,25 @@ Other requirements:
...
-[1] If `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` or `SPLUNK_METRICS_INGEST_URL` is set, takes precedence over the `SPLUNK_REALM` setting.
-
-
+Apart from standard set of configuration properties for instrumentation libraries based on OpenTelemetry, serverless MUST honour the following:  
+
+| Name                | Default value        | Description                                            |
+| ------------------- | -------------------  | ------------------------------------------------------ |
+| `SPLUNK_REALM`      | none                 | Splunk Observability Cloud ingest realm [1]            |
+
+- [1] Either `SPLUNK_REALM` or relevant traces exporter endpoint property and `SPLUNK_METRICS_INGEST_URL` MUST be set.
```
mateuszrzeszutek, line 176: Wasn't it supposed to be `SPLUNK_METRICS_ENDPOINT`?

specification/configuration.md
```diff
@@ -163,6 +163,27 @@ Other requirements:
 - RUM library MUST use the Zipkin v2 JSON span exporter by default
 - RUM library MUST limit the number of sent spans to 100 in a 30 second window per `component` attribute value
 
+#### Serverless 
+
+Current serverless offering is composed of separate metrics (non OpenTelemetry) and traces (OpenTelemetry) solutions. By default, the serverless components send data directly to Splunk Observability Cloud (direct ingest).
```
flands, line 168:
```suggestion
By default, serverless components MUST send data directly to Splunk Observability Cloud (direct ingest). 
```

specification/configuration.md
```diff
@@ -163,6 +163,27 @@ Other requirements:
 - RUM library MUST use the Zipkin v2 JSON span exporter by default
 - RUM library MUST limit the number of sent spans to 100 in a 30 second window per `component` attribute value
 
+#### Serverless 
+
+Current serverless offering is composed of separate metrics (non OpenTelemetry) and traces (OpenTelemetry) solutions. By default, the serverless components send data directly to Splunk Observability Cloud (direct ingest). 
+   
+Apart from standard set of configuration properties for instrumentation libraries based on OpenTelemetry, serverless MUST honour the following:
```
flands, line 170: In addition to honoring, I think we should require token+realm or endpoint to be defined.

specification/configuration.md
```diff
@@ -163,6 +163,27 @@ Other requirements:
...
+- [1] Either `SPLUNK_REALM` or relevant traces exporter endpoint property and `SPLUNK_METRICS_INGEST_URL` MUST be set.
+
+    If `SPLUNK_REALM` is defined, both relevant traces and metrics exporter endpoints will have following values:
+    - traces (in case of `otlp`): `https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp` 
+    - traces (all other cases):`https://ingest.REALM.signalfx.com/v2/trace`
+    - metrics: `https://ingest.${SPLUNK_REALM}.signalfx.com`
+    
+    If relevant traces exporter endpoint property (eg `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` for `otlp`) or `SPLUNK_METRICS_INGEST_URL` is set, it takes precedence over the `SPLUNK_REALM` setting.
+    
+As there is no deployment phase in case of Serverless functions, if a required configuration property is missing, the serverless component MUST log an error.
```
flands, line 185: But the function MUST still run?

specification/configuration.md
```diff
@@ -163,6 +163,27 @@ Other requirements:
...
 
+#### Serverless 
+
+Current serverless offering is composed of separate metrics (non OpenTelemetry) and traces (OpenTelemetry) solutions. By default, the serverless components send data directly to Splunk Observability Cloud (direct ingest). 
+   
+Apart from standard set of configuration properties for instrumentation libraries based on OpenTelemetry, serverless MUST honour the following:  
+
+| Name                | Default value        | Description                                            |
+| ------------------- | -------------------  | ------------------------------------------------------ |
+| `SPLUNK_REALM`      | none                 | Splunk Observability Cloud ingest realm [1]            |
```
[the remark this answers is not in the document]
kubawach: Done.

specification/configuration.md
kubawach: Added remark.

specification/configuration.md
kubawach: Yes. Log but don't fail :)

specification/configuration.md
```diff
@@ -92,6 +92,7 @@ environment variables:
 | :------------------------------------:          | :--------------------------------------------------------:     |
 | `SPLUNK_ACCESS_TOKEN` ()                        | Access token added to exported data. [1]                       |
 | `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [2] |
+| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for non-OTel metrics data ingest.                     |
```
flands, line 95:
```suggestion
| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for metrics data ingest.                     |
```

flands: Just to confirm @mateuszrzeszutek this is what splunk-otel-java uses?

kubawach: In the name of Mateusz - yes it is.

PR #138, opened 2021-10-20
spec proposal - additional OTLP exporter values

kubawach:


specification/configuration.md
```diff
@@ -133,7 +133,10 @@ beyond the OpenTelemetry specification exist.
 - Zipkin exporter
   - Distribution MUST NOT list Zipkin exporter as supported (not supported by Smart Agent)
 - `OTEL_TRACES_EXPORTER`
-  - Non-RUM distribution MUST default to `otlp` over gRPC with an endpoint of `localhost:4317`
+  - Non-RUM distribution MUST support following, additional values:
+    - `otlp-grpc` as an alias for `OTEL_EXPORTER_OTLP_PROTOCOL` value `grpc`
+    - `otlp-http` as an alias for `OTEL_EXPORTER_OTLP_PROTOCOL` value `http/protobuf`
```
pellared, line 138:
```suggestion
    - `otlp-grpc` as an alias for `OTEL_TRACES_EXPORTER=otlp` and  `OTEL_EXPORTER_OTLP_PROTOCOL=grpc`
    - `otlp-http` as an alias for `OTEL_TRACES_EXPORTER=otlp` and  `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`
```

specification/configuration.md
```diff
@@ -133,7 +133,10 @@ beyond the OpenTelemetry specification exist.
 - Zipkin exporter
   - Distribution MUST NOT list Zipkin exporter as supported (not supported by Smart Agent)
 - `OTEL_TRACES_EXPORTER`
-  - Non-RUM distribution MUST default to `otlp` over gRPC with an endpoint of `localhost:4317`
+  - Non-RUM distribution MUST support following, additional values:
```
pellared, line 136: I think that `MUST` in this context introduced a breaking-change. Maybe it would be to change it to `SHOULD`? Otherwise, I think we should bump the major version of the spec.

kubawach: I prefer MUST here, as SHOULD won't introduce a consistency between libraries. Then the docs will be much more messy and overall APM harder to use. Say:
- in python set single variable 
- in java set two, with different values
If we want to make the experience better for end-users, we need to have it supported everywhere.

pellared:
@owais WDYT about adding support for `SPLUNK_REALM` and `SPLUNK_ACCESS_TOKEN` instead so that the user does not need to specify `OTEL_TRACES_EXPORTER` at all?

kubawach:
> Assumption: #140 is accepted and a PR merged.
> 
> Would then this PR would still be worth the cost (as the instrumentation repositories would need to implement it 😄 )?

I think so. 
REALM support is from the same category (making end-user life easier) but in a different way - avoiding potential issues with specifying full endpoint URLs. 
Proposed exporter IDs would serve the same purpose. Even in our community I see sometimes problems with understanding that an exporter can have a protocol and there is a need to set both to get proper results. I also think it should be pretty minimal effort to implement.

iNikem:
Just to be sure, I want to point out that Otel spec defaults to otlp/http, not otlp/grpc: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol

mateuszrzeszutek:
> If no configuration is provided the default transport SHOULD be `http/protobuf` unless SDKs have good reasons to choose `grpc` as the default (e.g. for backward compatibility reasons when `grpc` was already the default in a stable SDK release).

Java uses `grpc` as the default by the way

pellared:
> Java uses `grpc` as the default by the way

As far as I remember .NET and Go as well.

BTW:

> If no configuration is provided the default transport SHOULD be `http/protobuf`
> unless SDKs have good reasons to choose `grpc` as the default (e.g. for backward
> compatibility reasons when `grpc` was already the default in a stable SDK
> release).

kubawach:
Anyway, the conversation should be around following question: "Do we believe that adding new exporter values - `otlp-grpc` and `otlp-http` is worth the effort. That is if extra engineering effort will be offset by lowering friction to end users".

pellared:
> Anyway, the conversation should be around following question: "Do we believe that adding new exporter values - `otlp-grpc` and `otlp-http` is worth the effort. That is if extra engineering effort will be offset by lowering friction to end users".

Depends how often users would really need to use it. If almost never I do not think is worth the effort. I believe that https://github.com/signalfx/gdi-specification/issues/140 would cover ~95% of the cases and would be easier to use.

specification/configuration.md
```diff
@@ -133,7 +133,10 @@ beyond the OpenTelemetry specification exist.
...
+    - `otlp-grpc` as an alias for `OTEL_TRACES_EXPORTER=otlp` and  `OTEL_EXPORTER_OTLP_PROTOCOL=grpc`
+    - `otlp-http` as an alias for `OTEL_TRACES_EXPORTER=otlp` and  `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`
```
MrAlias, line 138: This seems to be introducing an alternate way to specify the exporter than what is present in OpenTelemetry. Here it is specified as a single value, in OpenTelemetry it is specified as two. They both achieve the same thing though. By introducing an alternate way to specify this exporter Splunk's use of OpenTelemetry will diverge.

Instead of creating our own way to define these exporters, I think the better approach is to say projects need to support the `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable. That way we will remain in-line with OpenTelemetry and still achieve what this PR is aiming to accomplish.

PR #139, opened 2021-10-20
otel_log_level (common debug) proposal

kubawach:


pellared (commented):
I think that this PR would only add noise as I think that it is obvious that if something is defined in [OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration) then we should support it.

The only addition is that the spec only mentions `info` and here we are having `INFO`, `DEBUG`, `TRACE`. Maybe we should rather make a PR to the OTel spec with value proposals.

specification/configuration.md
```diff
@@ -135,7 +135,11 @@ beyond the OpenTelemetry specification exist.
 - `OTEL_TRACES_EXPORTER`
   - Non-RUM distribution MUST default to `otlp` over gRPC with an endpoint of `localhost:4317`
   - Non-RUM distribution MUST offer `jaeger-thrift-splunk` that defaults to `http://127.0.0.1:9080/v1/trace`
-
+- `OTEL_LOG_LEVEL`
+  - Distribution SHOULD support `DEBUG` setting. With this value all relevant troubleshooting data should be logged.
```
pellared, line 139: AFAIK supporting this would involve hacking in a most of SDKs

owais: https://github.com/open-telemetry/opentelemetry-specification/issues/920

specification/configuration.md
```diff
@@ -135,7 +135,11 @@ beyond the OpenTelemetry specification exist.
 - `OTEL_TRACES_EXPORTER`
   - Non-RUM distribution MUST default to `otlp` over gRPC with an endpoint of `localhost:4317`
   - Non-RUM distribution MUST offer `jaeger-thrift-splunk` that defaults to `http://127.0.0.1:9080/v1/trace`
-
+- `OTEL_LOG_LEVEL`
+  - Distribution SHOULD support `DEBUG` setting. With this value all relevant troubleshooting data should be logged. 
+  - Distribution MAY support additional levels:
+    - `INFO` - coarser-grained informational events than the `DEBUG`.
+    - `TRACE` - finer-grained informational events than the `DEBUG`.
```
mateuszrzeszutek, line 142: If we're going to use `OTEL_LOG_LEVEL` (which is a good idea, since it's stable in the otel spec) I believe we should wait for the OTel spec issues to get resolved before deciding on anything here.

mateuszrzeszutek (changes requested):

kubawach:
> > It is obvious, but for non-OTel (yes, we have non-OTel components) it is not. Hence this PR.
> 
> I wasn't aware of that - What kind of non-OTel components are we expecting(or are already) to follow the GDI spec?
>

Currently - the metrics extension (based on SFx metrics) which is bundled with the Splunk OTel Lambda Layer. It has currently different config for logging which IMO is confusing for end-users. Hence this request.

specification/configuration.md
kubawach: Yeah, we need to wait for the values to be finalized in the spec. Then it will be a no-brainer ;)

PR #145, opened 2021-11-03
Add packaging specification information

flands:
Proposal to get Helm chart to 1.0

### DIFF
--- a/specification/versioning.md
+++ b/specification/versioning.md
@@ -52,6 +52,12 @@ component MUST clearly be marked as stable.
 All components MAY become stable together, or MAY transition to
 stability component-by-component.
 
+Different packaging options may exist for GDI repositories. Packaging MAY have
+its own GDI repository. For example, the Helm chart for the Data Collector
+exists in its own repository. Packaging that exists in a dedicated GDI
+repositories MAY go stable even if components contained within the packaging
+are not stable.
+
 Once a component is marked as stable, the following rules MUST apply
 until the end of that component’s existence.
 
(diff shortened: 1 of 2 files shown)


pellared:
While I understand its purpose I think this proposal can bring confusion. E.g. who will decide if a dependency can be not stable? E.g. I see npm packages as a packaging/deployment for JS instrumentation and I do not think we want to go stable before OTLP exporter goes stable.

How about describing that dependencies `MUST` be stable + a list of exceptions.

flands:
> While I understand its purpose I think this proposal can bring confusion. E.g. who will decide if a dependency can be not stable? E.g. I see npm packages as a packaging/deployment for JS instrumentation and I do not think we want to go stable before OTLP exporter goes stable.
> 
> How about describing that dependencies `MUST` be stable + a list of exceptions.

I am trying to try a distinction between dependencies and components, but they are similar. My point here is that GDI packaging repositories may go stable independently from binaries they contain. Right now, this applies to Helm, Operator, Ansible, Puppet, and Heroku.

specification/versioning.md
```diff
@@ -52,6 +52,12 @@ component MUST clearly be marked as stable.
 All components MAY become stable together, or MAY transition to
 stability component-by-component.
 
+Different packaging options may exist for GDI repositories. Packaging MAY have
+its own GDI repository. For example, the Helm chart for the Data Collector
+exists in its own repository. Packaging that exists in a dedicated GDI
+repositories MAY go stable even if components contained within the packaging
```
pellared, line 58:
```suggestion
repository MAY go stable even if components contained within the packaging
```
 
(not sure as I am not a native speaker)

pellared (approved):

dmitryax (approved):

MrAlias (approved):

Issue #133, opened 2021-10-04
GA request for splunk-otel-android


jkwatson:
**Which GDI repository do you wish to GA?**
[splunk-otel-android](https://github.com/signalfx/splunk-otel-android)

**Does the repository follow the latest tagged minor release in GDI specification?**

  - [x] Has an appropriate [maintainers team](https://github.com/signalfx/gdi-specification/blob/v1.1.0/specification/repository.md#teams). - https://github.com/orgs/signalfx/teams/gdi-java-maintainers
  - [x] [Permissions](https://github.com/signalfx/gdi-specification/blob/v1.1.0/specification/repository.md#permissions) set correctly.
  - [x] [Branch protection](https://github.com/signalfx/gdi-specification/blob/v1.1.0/specification/repository.md#branch-protection) in place.
  - [x] [Dependencies](https://github.com/signalfx/gdi-specification/blob/v1.1.0/specification/repository.md#dependencies) appropriately locked down : [link](https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/build.gradle#L42-L64).
  - [x] [GitHub Applications](https://github.com/signalfx/gdi-specification/blob/v1.1.0/specification/repository.md#github-applications) set up per spec. 
  - [x] [Required Files](https://github.com/signalfx/gdi-specification/blob/v1.1.0/specification/repository.md#required-files) in place.
    - [x] [CHANGELOG.md](https://github.com/signalfx/splunk-otel-android/blob/main/CHANGELOG.md)
    - [x] [CODE_OF_CONDUCT.md](https://github.com/signalfx/splunk-otel-android/blob/main/CODE_OF_CONDUCT.md)
    - [x] [CONTRIBUTING.md](https://github.com/signalfx/splunk-otel-android/blob/main/CONTRIBUTING.md)
    - [x] [.github/CODEOWNERS](https://github.com/signalfx/splunk-otel-android/blob/main/.github/CODEOWNERS)
    - [x] [LICENSE](https://github.com/signalfx/splunk-otel-android/blob/main/LICENSE)
    - [x] [README.md](https://github.com/signalfx/splunk-otel-android/blob/main/README.md)
      - [x] Build status badge
      - [x] [Getting started](https://github.com/signalfx/splunk-otel-android#getting-started)
      - [x] [Troubleshooting](https://github.com/signalfx/splunk-otel-android#troubleshooting)
[...]

PR #146, opened 2021-11-03
Clarify RUM beaconEndpoint value enforcement

MrAlias:
Some systems, Android, cannot enforce this configuration value at the library level. It needs to be enforced at the application level. Therefore, this part of the specification is unimplementable by the Android RUM library. This adds an exception to this requirement to allow the Android RUM library to be compliant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 #### Bug fixes
 
 - Properly specify configuration statuses for RUM and serverless.
+- Clarify requirement for RUM `beaconEndpoint` configuration value to be HTTPS
+  only on systems that allow such enforcement.
 
 ### Versioning
 
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -180,10 +180,11 @@ properties:
 - [1] Application name, authentication token and either realm or the beacon URL
   MUST be provided by the user. If any of these is missing, the RUM
   instrumentation library MUST fail to start.
-- [2] Implementations MUST enforce by default that `beaconEndpoint`s are https
-  only, and reject/fail to start otherwise. Implementations MAY offer an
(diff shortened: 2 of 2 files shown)


MrAlias:
Issue identified here: https://github.com/signalfx/gdi-specification/issues/133#issuecomment-959809133

MrAlias:
@jkwatson please take a look (not sure why I cannot request you as a reviewer).

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
 #### Bug fixes
 
 - Properly specify configuration statuses for RUM and serverless.
+- Clarify requirement for RUM `beaconEndpoint` configuration value to be HTTPS
+  only on systems that allow such enforcement.
```
jkwatson, line 17: So...to clarify a bit... In the Android library, we *could* add an additional check that required the user to set things to https, and allow that to be overridden by a the config option. But, it wouldn't actually do anything useful, since whether to allow bare `http` client requests (or not, which is the default behavior) is controlled by the Android app manifest itself.

MrAlias: I'm a bit out of my depth on this one. Does this mean we need to change the language of this PR to better address the situation?

jkwatson: no, I think the wording there is fine. I guess I just wanted to highlight the nuance that we *could* enforce that the endpoint that people provided was https, but it wouldn't really have any meaningful impact. Or, perhaps more importantly, if we gave them an override, it would be misleading, since they would still have to configure their app to use insecure http calls for it to do anything.

flands (approved):

mateuszrzeszutek (approved):

owais (approved):

PR #147, opened 2021-11-03
Add data collection configuration for Kubernetes

flands:
Also add language around Splunk Observability versus Platform

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -95,7 +95,9 @@ environment variables:
 While Kubernetes supports container technology that can be configured using
 environment variables, package management solutions such as Helm charts and
 Operators require YAML-based configuration. As a result, Kubernetes package
-management solutions MUST support the following YAML configuration options:
+management solutions MUST support the YAML configuration options specified below.
+
+> Any option description listed as REQUIRED means a value for the option MUST be specified.
 
 - `clusterName` ()             : [REQUIRED] Name of the cluster.
 - `cloudProvider` ()           : Where Kubernetes is deployed.
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -111,7 +111,7 @@ management solutions MUST support the following YAML configuration options:
   - `enabled` (`true`)         : Whether k8sClusteReceiver is deployed. Ignored if `metricsEnabled` is `false`.
   - `config` ()                : Updates configuration. Non-list options merged, list options override.
 
-In addition, at least one of the following sections MUST be specified:
+In addition, at least one of the the following configuration groups, `splunkObservability` or `splunkPlatform`, MUST be specified:
 
 - `splunkObservability`
   - `accessToken` ()           : [REQUIRED] Access token added to exported data.
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -6,9 +6,7 @@ One or more configuration variables may be needed to properly configure GDI
 repositories. Components that can be configured with environment variables MUST
 support configuration of these variables using environment variables. Any
 component that cannot be configured with environment variables MUST support
-configuration of these variables using an alternate method and that method MUST
-be defined in a section below (see the [Real User Monitoring Libraries
-section](#real-user-monitoring-libraries) as an example). Any component MAY
+configuration of these variables using an alternate method. Any component MAY
 support configuration of these variables by additional methods.
 
 GDI repositories MUST adopt stable and SHOULD adopt experimental configuration
@@ -32,15 +30,18 @@ release. This change MAY result in a breaking change so caution should be
 exhibited when considering repository-specific configuration variables.
 
 Splunk-specific configuration variables defined in the GDI specification MUST
-be prefixed with `SPLUNK_`. If a Splunk-specific configuration variable is
-declared as stable in the GDI specification and later the OpenTelemetry
-specification declares a similar variable as stable, the GDI specification
-MUST adopt the OpenTelemetry configuration variable and SHOULD mark the GDI
-specification configuration variable as deprecated by the next minor release.
-In addition to defining Splunk-specification configuration variables, the GDI
-specification MAY require specific OpenTelemetry configuration variables be
-supported. If it does, the GDI specification MAY require certain values be
-supported including a specific default value.
+be prefixed with `SPLUNK_`. Furthermore, configuration specific to Splunk
+Observability Cloud MUST be prefixed with `SPLUNK_OBSERVABILITY_` and to Splunk
+Enterprise or Splunk Cloud MUST be prefixed with `SPLUNK_PLATFORM_`. If a
+Splunk-specific configuration variable is declared as stable in the GDI
+specification and later the OpenTelemetry specification declares a similar
+variable as stable, the GDI specification MUST adopt the OpenTelemetry
+configuration variable and SHOULD mark the GDI specification configuration
+variable as deprecated by the next minor release. In addition to defining
+Splunk-specification configuration variables, the GDI specification MAY require
+specific OpenTelemetry configuration variables be supported. If it does, the
+GDI specification MAY require certain values be supported including a specific
+default value.
 
 Whenever a configuration variable changes its name, a stable GDI repository
 (version >= 1.0) MUST support both old and new names until the next major
@@ -69,11 +70,11 @@ chooses what to install limiting the configuration options.
 It MUST be possible to configure a Data Collector instance using the following
 environment variables:
 
-| Name (default value)     | Description                                        |
-| :-------------------:    | :-----------------------------------:              |
-| `SPLUNK_ACCESS_TOKEN` () | Access token added to exported data. [1][2]        |
-| `SPLUNK_CONFIG` ()       | Configuration file to use. [1]                     |
-| `SPLUNK_REALM` ()        | Realm configured for the exporter endpoint. [1][2] |
+| Name (default value)          | Description                                        |
+| :-------------------:         | :-----------------------------------:              |
+| `SPLUNK_ACCESS_TOKEN` ()      | Access token added to exported data. [1][2]        |
+| `SPLUNK_CONFIG` ()            | Configuration file to use. [1]                     |
+| `SPLUNK_REALM` ()             | Realm configured for the exporter endpoint. [1][2] |
 
 - [1]: Either `SPLUNK_ACCESS_TOKEN` and `SPLUNK_REALM` MUST be defined or
   `SPLUNK_CONFIG` MUST be defined. If `SPLUNK_ACCESS_TOKEN` and `SPLUNK_REALM`
@@ -88,6 +89,50 @@ environment variables:
   [`access_token_passthrough`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/signalfxreceiver#configuration)).
   If `SPLUNK_CONFIG` is defined then these options are not required but MAY be
   defined.
+
+### Kubernetes
+
+While Kubernetes supports container technology that can be configured using
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
+
+- `splunkCommon`
+  - `clusterName` ()          : [REQUIRED] Name of the cluster.
+  - `distro` ()               : Which distribution of Kubernetes is deployed.
+  - `environment` ()          : Name of the environment; if not defined then skipped.
+  - `logsEngine` (`fluentd`)  : How to collect logs.
+  - `modes`
+    - `agent`                 : Deployed as a DaemonSet.
+      - `config` ()           : Updates configuration. Non-list options merged, list options override.
+    - `k8sClusterReceiver`    : Deployed as a standalone service.
+      - `config` ()           : Updates configuration. Non-list options merged, list options override.
+    - `gateway`               : Deployed as a clustered Service.
+      - `enabled` (`false`)   : Whether mode is deployed.
+      - `config` ()           : Updates configuration. Non-list options merged, list options override.
+  - `provider` ()             : Where Kubernetes is deployed.
+- `splunkObservability`
+  - `accessToken` ()          : [REQUIRED] Access token added to exported data.
+  - `logsEnabled` (`false`)   : Whether logs are collected and sent.
+  - `metricsEnabled` (`true`) : Whether metrics are collected, received, and sent.
+  - `realm` ()                : [REQUIRED] Realm configured for the exporter endpoint.
+  - `tracesEnabled` (`true`)  : Whether traces can be received and sent.
+- `splunkPlatform`
+  - `endpoint` ()             : [REQUIRED] Where to send exported data.
+  - `logsEnabled` (`false`)   : Whether logs are collected and sent.
+  - `metricsEnabled` (`true`) : Whether metrics are collected, received, and sent.
+  - `token` ()                : [REQUIRED] Token added to exported data.
+
+In addition, the following Kubernetes secret configuration options MUST be
+supported:
+
+- Splunk Observability
+  - `splunk_observability_access_token`
+- Splunk Platform
+  - `splunk_platform_hec_token`
+  - `hec_client_cert`
+  - `hec_client_key`
+  - `hec_ca_file`
 
 ## Instrumentation Libraries
 
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -95,9 +95,11 @@ environment variables:
 While Kubernetes supports container technology that can be configured using
 environment variables, package management solutions such as Helm charts and
 Operators require YAML-based configuration. As a result, Kubernetes package
-management solutions MUST support the YAML configuration options specified below.
+management solutions MUST support the YAML configuration options specified
+below.
 
-> Any option description listed as REQUIRED means a value for the option MUST be specified.
+> Any option description listed as REQUIRED means a value for the option MUST
+> be specified.
 
 - `clusterName` ()             : [REQUIRED] Name of the cluster.
 - `cloudProvider` ()           : Where Kubernetes is deployed.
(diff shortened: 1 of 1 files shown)


specification/configuration.md
```diff
@@ -89,6 +90,50 @@ environment variables:
...
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
+
+- `splunkCommon`
+  - `clusterName` ()          : [REQUIRED] Name of the cluster.
+  - `distro` ()               : Which distribution of Kubernetes is deployed.
+  - `environment` ()          : Name of the environment; if not defined then skipped.
+  - `logsEngine` (`fluentd`)  : How to collect logs.
+  - `modes`
```
dmitryax, line 105: I don't agree that this is a `mode`. Helm can install the connector in many modes, so it's rather a component that can be enabled or disabled than a mode as in other single-component packaging solutions.

specification/configuration.md
```diff
@@ -89,6 +90,50 @@ environment variables:
...
   defined.
 
+### Kubernetes
+
+While Kubernetes supports container technology that can be configured using
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
+
+- `splunkCommon`
```
dmitryax, line 100: Assuming we target helm initially in this specification. The existing configuration interface in the helm chart tend to be flat and don't use grouping unless it's necessary following helm best practices https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values

Do we need this particular grouping under `splunkCommon`? Can we assume any flat configuration other than `splunk<Observability|Platform>` considered to be common?

specification/configuration.md
```diff
@@ -89,6 +90,50 @@ environment variables:
...
+In addition, the following Kubernetes secret configuration options MUST be
+supported:
+
+- Splunk Observability
+  - `splunk_observability_access_token`
+- Splunk Platform
+  - `splunk_platform_hec_token`
+  - `hec_client_cert`
+  - `hec_client_key`
+  - `hec_ca_file`
```
dmitryax, line 135: I think we need to add `splunk_platform_` prefix to all of these. Splunk Observability doesn't support ,them but they would be different anyway

specification/configuration.md
```diff
@@ -89,6 +90,50 @@ environment variables:
...
+### Kubernetes
+
+While Kubernetes supports container technology that can be configured using
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
+
+- `splunkCommon`
+  - `clusterName` ()          : [REQUIRED] Name of the cluster.
+  - `distro` ()               : Which distribution of Kubernetes is deployed.
```
dmitryax, line 102: `platform` is needed here as well I believe

specification/configuration.md
```diff
@@ -89,6 +90,50 @@ environment variables:
...
+  - `environment` ()          : Name of the environment; if not defined then skipped.
+  - `logsEngine` (`fluentd`)  : How to collect logs.
+  - `modes`
+    - `agent`                 : Deployed as a DaemonSet.
+      - `config` ()           : Updates configuration. Non-list options merged, list options override.
+    - `k8sClusterReceiver`    : Deployed as a standalone service.
+      - `config` ()           : Updates configuration. Non-list options merged, list options override.
```
dmitryax, line 109: Both of them need `enabled` option as well. For example, they both can be disabled in order to deploy gateway-only mode

specification/configuration.md
flands: In our docs today, we call it a mode: https://github.com/signalfx/splunk-otel-collector/blob/main/docs/agent-vs-gateway.md. I think we should be consistent. How about `deployment`?

specification/configuration.md
flands: The guidance in the link appears to go both way. For example: `When there are a large number of related variables, and at least one of them is non-optional, nested values may be used to improve readability.` My concern is that there are a lot of additional options and lack of nesting for confusing for me.

What if we do:

```
splunk:
  environment:
  observability:
    accessToken:
  platform:
    token:
```

specification/configuration.md
flands: well LO supports hec... maybe just prefix with `splunk_`?

specification/configuration.md
flands: We need platform, provider, and distro?? Are all required? Do we have a default for any?

specification/configuration.md
flands: Ah, makes sense.

specification/configuration.md
dmitryax: Can we keep "mode" term for single-component deployments and use something else like "component" multi-component deployments?

"deployment" doesn't seem to be a good choice to me because it can be confused with k8s deployment resource, and "agent" is a daemonset not deployment.

Looking at the customer experience so far the most common usage is adjusting recourses and adding extra components to the default agent's config. With the current helm chart, their custom helm values file would look like this:

```yaml
splunkObservability:
  accessToken: xxx
  realm: us0
otelAgent:
  resources:
    limits:
      cpu: 500m
      memory: 1Gi
  config:
    receivers:
      smartagent/docker-container-stats:
        type: docker-container-stats
        dockerURL: unix:///hostfs/var/run/docker.sock
    service:
      pipelines:
        metrics:
          receivers:
            - smartagent/docker-container-stats
            - hostmetrics
            - kubeletstats
            - otlp
            - receiver_creator
            - signalfx
```

It's already pretty deep nesting for agent configuration changes, and I heard about many cases when users confuse the layers. If we add 2 more levels on top of it (`splunkCommon` -> `modes`), I believe it'll make user experience worse.

And many of configurations options specific to helm chart release and kubernetes resources are on the top level because they already have established naming patterns in helm ecosystem, examples:

```
nameOverride: ""
fullnameOverride: ""
serviceAccount:
secret:
rbac:
```

Those options are pretty well known to those who used other helm chart before. So I think we should keep them at the top level. 

To summarize, I would really recommend to keep the components groups at the top level as well. May be we can prefix them with some common word, something like `componentAgent`, `componentGateway` and `componentK8sClusterReceiver`? I took `otel` prefix because we still have `fluent` as another component, so to keep the distinction now we have `otelAgent`, `otelCollector`, `otelK8sClusterReceiver` and `fluentd`. And for me it looks better than other alternatives, assuming `otelCollector` is renamed to `otelGateway`

specification/configuration.md
dmitryax: LO supports hec, but can we assume that the same certificates always must be used for both backends?

specification/configuration.md
dmitryax: Oh sorry, I mean `provider`, not `platform`. We need `provider` and `distro`, both of them are optional

specification/configuration.md
owais: +1 this is what the operator does as well i.e, it accepts "common" fields at the root level and then has nested levels for different components e.g, 

```yaml
spec:
  clusterName: <MY_CLUSTER_NAME>
  realm: <SPLUNK_REALM>
  agent:
     ...
  gateway:
     ...
```

https://github.com/signalfx/splunk-otel-collector-operator/blob/main/docs/splunkotelagent_cr_spec.md

specification/configuration.md
owais: FWIW "mode" does sound a bit confusing to me as well. It sounds as if I can only pick one of the three modes especially if I'm using a single/common config file to deploy multiple agents.

specification/configuration.md
```diff
@@ -89,6 +90,53 @@ environment variables:
...
+### Kubernetes
+
+While Kubernetes supports container technology that can be configured using
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
+
+- `clusterName` ()            : [REQUIRED] Name of the cluster.
+- `environment` ()            : Name of the environment; if not defined then skipped.
+- `distro` ()                 : Which distribution of Kubernetes is deployed.
```
dmitryax, line 102: Maybe full name `distribution`?

specification/configuration.md
```diff
@@ -89,6 +90,53 @@ environment variables:
...
+- `provider` ()               : Where Kubernetes is deployed.
```
dmitryax, line 103: Don't have strong preference but maybe we should consider `cloudProvider`?

dmitryax (approved):
LGTM

specification/configuration.md
```diff
@@ -89,6 +90,52 @@ environment variables:
...
   defined.
 
+### Kubernetes
+
+While Kubernetes supports container technology that can be configured using
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
+
+- `clusterName` ()             : [REQUIRED] Name of the cluster.
```
dmitryax, line 100: This is actually required only if Observability destination is enabled. I think we should mention that

specification/configuration.md
```diff
@@ -89,6 +90,52 @@ environment variables:
   If `SPLUNK_CONFIG` is defined then these options are not required but MAY be
   defined.
 
+### Kubernetes
```
MrAlias, line 93:
```suggestion
### Kubernetes Package Management Solutions
```

specification/configuration.md
```diff
@@ -89,6 +90,52 @@ environment variables:
...
+  - `enabled` (`true`)         : Whether agent is deployed.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+- `gateway`                    : Deployed as a clustered Service and receives data from agent.
+  - `enabled` (`false`)        : Whether gateway is deployed.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+- `clusterReceiver`            : Deployed as a single replica deployment and collects Kubernetes API cluster and event telemetry.
+  - `enabled` (`true`)         : Whether k8sClusteReceiver is deployed. Ignored if `metricsEnabled` is `false`.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+
+In addition, at least one of the following sections MUST be specified:
```
MrAlias, line 114: "following sections" is a bit vague, it can be interpreted as a paragraph or markdown section as well as a level in the following list. Can we nail this down?

```suggestion
In addition, at least one of the the following configuration groups, `splunkObservability` or `splunkPlatform`, MUST be specified:
```

specification/configuration.md
```diff
@@ -89,6 +90,52 @@ environment variables:
...
+Finally, the following Kubernetes secret configuration options MUST be
+supported:
+
+- Splunk Observability
+  - `splunk_observability_access_token`
+- Splunk Platform
+  - `splunk_platform_hec_token`
+  - `splunk_platform_hec_client_cert`
+  - `splunk_platform_hec_client_key`
+  - `splunk_platform_hec_ca_file`
```
MrAlias, line 137: Should these be interwoven with the above list? They seem subsets of the above configuration options, and potentially duplicates (i.e. `splunk_observability_access_token` looks very similar to `accessToken` in the `splunkObservability` list element)

Also, should they include descriptions and requirement levels?

specification/configuration.md
MrAlias: Does it make sense to the authors of Kuberntetes package management solutions what required means? Do we need to define behavior for them when a required variable is not provided?

specification/configuration.md
flands: This is specific for k8s secrets. @dmitryax if you specify this do you need to specify say `accessToken` above?

specification/configuration.md
```diff
@@ -89,6 +90,52 @@ environment variables:
   If `SPLUNK_CONFIG` is defined then these options are not required but MAY be
   defined.
 
+### Kubernetes Package Management Solutions
+
+While Kubernetes supports container technology that can be configured using
+environment variables, package management solutions such as Helm charts and
+Operators require YAML-based configuration. As a result, Kubernetes package
+management solutions MUST support the following YAML configuration options:
```
flands, line 98:
```suggestion
management solutions MUST support the YAML configuration options specified below.

> Any option description listed as REQUIRED means a value for the option MUST be specified.
```

specification/configuration.md
```diff
@@ -89,6 +90,52 @@ environment variables:
...
+  - `enabled` (`true`)         : Whether agent is deployed.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+- `gateway`                    : Deployed as a clustered Service and receives data from agent.
+  - `enabled` (`false`)        : Whether gateway is deployed.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+- `clusterReceiver`            : Deployed as a single replica deployment and collects Kubernetes API cluster and event telemetry.
+  - `enabled` (`true`)         : Whether k8sClusteReceiver is deployed. Ignored if `metricsEnabled` is `false`.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+
+In addition, at least one of the the following configuration groups, `splunkObservability` or `splunkPlatform`, MUST be specified:
```
flands, line 114:
```suggestion
In addition, at least one of the below configuration groups, 
`splunkObservability` or `splunkPlatform`, MUST be specified.

> Any option description listed as REQUIRED means a value for the option MUST be specified.
```

specification/configuration.md
flands: Addressed your comment @MrAlias . @dmitryax I think we should be consistent -- either both require or both don't.

specification/configuration.md
dmitryax: > This is specific for k8s secrets. @dmitryax if you specify this do you need to specify say accessToken above?
No, the secrets by default are set from yaml configuration, e.g. `splunkObservability.accessToken` goes to `splunk_observability_access_token`. But, if user wants to specify secrets by themselves, they can set them directly (`splunk_observability_access_token`) -- in that case `splunkObservability.accessToken` can be omitted.

specification/configuration.md
dmitryax: Ok, I'll check on this with SCK team

specification/configuration.md
MrAlias: > > This is specific for k8s secrets. @dmitryax if you specify this do you need to specify say accessToken above?
> 
> No, the secrets by default are set from yaml configuration, e.g. `splunkObservability.accessToken` goes to `splunk_observability_access_token`. But, if user wants to specify secrets by themselves, they can set them directly (`splunk_observability_access_token`) -- in that case `splunkObservability.accessToken` can be omitted.

Do we need an additional `Kubernetes` section? It seems like this section might be serving two topics? Also, would we not prefer environment variables for native kube configuration of the Splunk observability access token?

flands: I updated the previous section to hopefully clarify.

specification/configuration.md
```diff
@@ -89,6 +90,64 @@ environment variables:
...
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+- `clusterReceiver`            : Deployed as a single replica deployment and collects Kubernetes API cluster and event telemetry.
+  - `enabled` (`true`)         : Whether k8sClusteReceiver is deployed. Ignored if `metricsEnabled` is `false`.
+  - `config` ()                : Updates configuration. Non-list options merged, list options override.
+
+In addition, at least one of the below configuration groups,
+`splunkObservability` or `splunkPlatform`, MUST be specified.
+
+> Any option description listed as REQUIRED means a value for the option MUST
+> be specified. Both `accessToken` and `token` values are stored as Kubernetes
```
dmitryax, line 122: Should we clarify that `[REQUIRED]` here means that the option is required only within a particular group?

dmitryax (approved):

specification/configuration.md
```diff
@@ -89,6 +90,64 @@ environment variables:
...
+`splunkObservability` or `splunkPlatform`, MUST be specified.
+
+> Any option description listed as REQUIRED means a value for the option MUST
+> be specified in that section. Both `accessToken` and `token` values are
+> stored as Kubernetes secrets. The secret key names are specified after this
+> section. If Kubernetes secret keys are specified they will override the
+> values specified below. In short, `accessToken` and `token` are required only
+> if not specified in Kubernetes secrets.
+
+- `splunkObservability`
```
owais, line 128: nit: generally config like this should be namespaced/encapsulated so using `splunk` prefix is probably redundant. Can we call these just `observability` and `platform/enterprise`?

owais (commented):
Looks good. Left a nit-picky suggestion.

Q: Do we want to specify k8s API namespace(s) here? These are used whenever we install any custom resource definition in a k8s cluster. For example, the operator today uses `o11y.splunk.com` as the API namespace for Otel deployments. Do we want to prescribe this or out of scope for this PR? ref: https://github.com/signalfx/splunk-otel-collector-operator/blob/main/docs/splunkotelagent_cr_spec.md#splunkotelagent-custom-resource-specification

specification/configuration.md
MrAlias: It's been a while since I used helm, so I might be missing something, but as I remember it the namespace is defined by the operator when they install the chart. Meaning it is up to them to correctly define it as `splunk`, right? If so, it might make sense to keep the prefix, even though it is redundant, to ensure a clear scope to the variable.

MrAlias (approved):

specification/configuration.md
dmitryax: It's not really a prefix for Splunk specific configuration, but rather part of a destination configuration. Full names for the destinations/products are "Splunk Observability" and "Splunk Platform" (https://www.splunk.com/en_us/software.html). `observability` may be self-descriptive, but just `platform` is confusing I think. I would keep the full names to make it clear.

owais: That sounds reasonable and not a blocker for me but to answer the question,

> It's been a while since I used helm, so I might be missing something, but as I remember it the namespace is defined by the operator when they install the chart. Meaning it is up to them to correctly define it as splunk, right? 

My bad for using the term "namespace". I did not mean a k8s namespace but namespace in a general sense.

I was thinking of somethig like this:

```
k get splunk-agent

kind: SplunkOtelAgent
spec:
    clusterName: ...
    splunkObservability: ...
    splunkPlatform: ...
```

here repeating the word splunk in config feels unnecessary. 

Is this not the case for helm? Does splunk config get merged into an object that contains non-splunk config?

dmitryax: The configuration of the helm is applied to the whole chart release, it's on the root level, not a part of any object. There are some Splunk specific configs and Kubernetes specific configs in it. We already discussed extra grouping in this thread https://github.com/signalfx/gdi-specification/pull/147#discussion_r742334481. And I think we agreed that extra grouping like `splunkCommon` will make the user experience worse.

Looking at the same configuration in the `SplunkOtelAgent` k8s object, I think it's Ok to have some extra Splunk words in it given that `splunkPlatform` and `splunkObservability` refer to exact product names that IMO shouldn't be broken down.

owais (approved):

Issue #140, opened 2021-10-21
Support `SPLUNK_REALM` in instrumentation libraries


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Right now the "standard" instrumentation libraries do not support the `SPLUNK_REALM` configuration setting; it is supported by RUM & serverless libraries and the collector though.

We could introduce that setting to all instrumentation libraries; it could work in the following way:
* if `SPLUNK_REALM` is specified, the exporter endpoint will point to the ingest;
* if `SPLUNK_REALM` and the exporter endpoint config variable are both specified, the exporter setting takes precedence;
* when `SPLUNK_REALM` and exporter endpoint are not specified, the default endpoint values (pointing to collector) will be used.

pellared:
@owais WDYT of `none` instead of `local`? Just to have similarity to `OTEL_PROPAGATORS` and `OTEL_*_EXPORTER`.

Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md

PR #148, opened 2021-11-11
Add `SPLUNK_REALM` to instrumentation library configuration

flands:
Addresses #140

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -159,8 +159,9 @@ instance using the following environment variables:
 | Name (default value)                            | Description                                                    |
 | :------------------------------------:          | :--------------------------------------------------------:     |
 | `SPLUNK_ACCESS_TOKEN` ()                        | Access token added to exported data. [1]                       |
-| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [2] |
-| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for metrics data ingest.                     |
+| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for metrics data ingest.                              |
+| `SPLUNK_REALM` (`none`)                         | Which realm to send exported data. [2]                         |
+| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [3] |
 
 - [1]: Not user required if another system performs the authentication. For
   example, instrumentation libraries SHOULD send data to a locally running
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

specification/configuration.md
```diff
@@ -159,8 +159,9 @@ instance using the following environment variables:
 | Name (default value)                            | Description                                                    |
 | :------------------------------------:          | :--------------------------------------------------------:     |
 | `SPLUNK_ACCESS_TOKEN` ()                        | Access token added to exported data. [1]                       |
-| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [2] |
-| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for metrics data ingest.                     |
+| `SPLUNK_METRICS_ENDPOINT` ()                    | Endpoint for metrics data ingest.                              |
+| `SPLUNK_REALM` (`none`)                         | Which realm to send exported data. [2]                         |
```
owais, line 163: Other vars don't have a `none` default value and instead just default to nothing/empty string. Do we need a special default value ("none" string literal) for SPLUNK_REALM?

mateuszrzeszutek: I think we want to avoid using empty strings in general and use `none` in such cases in general: https://github.com/open-telemetry/opentelemetry-specification/pull/2102

It's also consistent with setting exporters, propagators (was that PR merged already?)

owais: That makes sense as it allows users to easily "unset" a variable that has a non-none default but in case the variable does not have a default, do we want to specify `none` as the default value or just say it has no default? I know it does not make a difference in implementation but when reading the spec it can be confusing when some values do not have a default value while others have `none` as default value while both having the same net effect.

owais (commented):
Forgot to submit pending comment 🤦

specification/configuration.md
pellared: > Other vars don't have a none default value and instead just default to nothing/empty string.

@owais, can you give an example of an enum configuration that works like that? If you find one, then I agree that `none` value make only sense when the default value is "not none".

pellared: > Other vars don't have a `none` default value and instead just default to nothing/empty string

@owais Can you give an example of such enum-type configuration, please?

owais: image
image
image
image


None of these have any default values. How do we decide when something does not have a default vs when something has default set to `none`?

pellared: `none` is for enum value ([more](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#enum-value). I think that only the Kubernetes configs could be enums but they look more like dummy strings at the moment - moreover there are experimental.

pellared (approved):

specification/configuration.md
owais: Yes, that makes sense and like I said this is not a hard blocker for me. Just that it can be confusing to me as an end user when I see a dozen options with some listing the default as `none` while others listing no default value and all of them behaving the same. As a consumer, I'd prefer a library to have uniform defaults/noop representations.

Feel free to resolve this if you'd like :)

owais (approved):

pellared:
@MrAlias PTAL

PR #149, opened 2021-11-11
Add behaviors

flands:
Refactor profiling into:

- Behaviors
- Configuration
- Semantic Conventions

Behavior is new to cover specifics of some subset. Its goal is to drive
consistent across GDI repositories.


specification/behaviors.md
```diff
@@ -0,0 +1,50 @@
...
+This document describes the specific behavior and conventions for Splunk
+instrumentation libraries that contain profiling features.
+
+### Call Stack Sampling
+
+An instrumentation library that has profiling capabilities MUST be able to
+sample call stacks at a fixed interval. This default interval MUST default to 1
+second.
+
+When a language runtime supports threading, stacks MUST be sampled across all
```
iNikem, line 16: unless we mention here that there is an exception described below in "filtering", we have a contradiction. "MUST be sampled across all threads" and "some threads SHOULD NOT be sampled"

specification/behaviors.md
```diff
@@ -0,0 +1,50 @@
...
+
+The profiler SHOULD NOT collect call stacks from undesirable threads. If this
+is not possible, the profiler SHOULD filter these out afterward and omit these
+call stacks from ingest.
+
+### Call Stack Span Association
+
+When a call stack is sampled during the execution of a span scope, the profiler
+MUST be able to associate the call stack to the span. This association SHOULD
+happen as close to the sampling point as feasible, but MAY occur later in a
```
iNikem, line 38: Why this closeness is important?

specification/semantic_conventions.md
```diff
@@ -76,3 +82,96 @@ Web RUM library MUST add the following attributes to all spans:
...
+`ResourceLogs` is the upstream protobuf data type. It MUST be populated with
+the attributes from the OpenTelemetry resource.
+
+### InstrumentationLibraryLogs
+
+Each `ResourceLogs` instance has an instance of `InstrumenationLibraryLogs`.
+For each `InstrumentationLibraryLogs` instance:
+
+- `name` - MUST be set to `otel.profiling`
+- `version` - MUST be set to `0.1.0`
```
iNikem, line 99: Why spec prescribes what looks like implementation version?

specification/behaviors.md
```diff
@@ -0,0 +1,50 @@
...
+processing pipeline.
```
iNikem, line 39: Inside the process or outside?

specification/semantic_conventions.md
```diff
@@ -76,3 +82,96 @@ Web RUM library MUST add the following attributes to all spans:
...
+```
+
+The thread states for other runtimes are currently undefined.
+
+The call stack details follow the thread state line. The call stack MUST have
+the top of the stack provided first and the bottom of the stack provided last.
+The format of each stack trace line is:
+
+```
+        at <namespace>.<function>(<file>:<lineno>)
```
iNikem, line 159: Should we just reuse https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md#stacktrace-representation ?

specification/behaviors.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+This document describes the specific behavior and conventions for Splunk
+instrumentation libraries that contain profiling features.
+
+### Call Stack Sampling
+
+An instrumentation library that has profiling capabilities MUST be able to
+sample call stacks at a fixed interval. This default interval MUST default to 1
```
laurit, line 13: java was changed to use 10 seconds

specification/behaviors.md
breedx-splk: This is a fair point, but I see the filtering as happening downstream of the sampling. The actual sampling mechanism may be more difficult to control (such as in the case of Java, where we can't control which threads JFR reports on). I don't think it's a contradiction.

specification/behaviors.md
breedx-splk: Not currently spec'd. Java is currently in process.

specification/behaviors.md
breedx-splk: It's probably really not that important. I think when I wrote this originally I thought that it would be helpful to encourage instrumentation to get the work done up front (eg. "Look it up and associate right when you get the stack trace, rather than deferring it")...but it's probably overspec'd tbh.

Our current JFR approach violates this `SHOULD` anyway.

specification/semantic_conventions.md
```diff
@@ -1,6 +1,6 @@
 # Semantic Conventions
 
-**Status**: [Stable](../README.md#versioning-and-status-of-the-specification)
+**Status**: [Mixed](../README.md#versioning-and-status-of-the-specification)
```
breedx-splk, line 3: We don't have a definition of "Mixed". Doesn't any non-`Stable` category here imply `Experimental`? I dunno.

specification/semantic_conventions.md
breedx-splk: The intent is to allow ingest/backend to be able to handle future changes to log body contents. So if `0.1.0` is just blob of text containing a stack trace but we version `0.2.0` to contains a more structured format (pprof or similar), the version could be used to detect this. 

Your question does make me wonder if `schemaUrl` isn't better suited to this purpose?

specification/semantic_conventions.md
breedx-splk: I spec'd it in more detail originally because the profiling backend has some assumptions.

specification/behaviors.md
```diff
@@ -0,0 +1,50 @@
...
+second.
```
breedx-splk, line 14:
```suggestion
sample call stacks at a fixed interval. This default interval MUST default to 10
seconds.
```

breedx-splk (approved):
I'm ok with this redistribution of content. It will take me a little getting used to, and I wonder if users (us!) will have a harder time finding what they are looking for, but I agree that this increases consistency. It's probably just an adjustment.

specification/semantic_conventions.md
iNikem: This is certainly not an intended use of `InstrumentationLibrary.version`. Schemas are probably better for this.

specification/semantic_conventions.md
flands: We do have a mixed: https://github.com/signalfx/gdi-specification/blob/bbc3951cabda7107a991f372a053d0f8685348ad/README.md#mixed

flands:
@iNikem @breedx-splk -- Nikita, it appears your comments are in regards to the existing content in the spec. Can I ask a separate issue/PR be created for this? My change simply redistributes the existing content (just looking for shortest path to merge this and to also separate changes in separate PRs)

flands:
> I'm ok with this redistribution of content. It will take me a little getting used to, and I wonder if users (us!) will have a harder time finding what they are looking for, but I agree that this increases consistency. It's probably just an adjustment.

This is a good point -- can you open an issue to track it? We might need some linking to address this or perhaps just need to refactor the whole thing.

iNikem:
@flands I don't object merging

PR #150, opened 2021-11-11
Require profiler stack sampling period to be configurable

breedx-splk:


specification/profiling.md
```diff
@@ -24,8 +24,10 @@ data.
 
 ## Call Stack Sampling
 
-An instrumentation library that has profiling capabilities MUST be able to sample call stacks
-at a fixed interval. This default interval MUST default to 1 second.
+An instrumentation library that has profiling capabilities MUST be able to sample 
+call stacks at a fixed interval. This interval MUST be user-configurable at startup 
+via the `SPLUNK_PROFILER_CALL_STACK_PERIOD` environment variable. When not explicitly
```
MrAlias, line 29: What unit is this value of the environment variable interpreted as? Seconds, milliseconds, microseconds ...

Should the unit be included in the variable name?

breedx-splk: In Java, the units are part of the value (`100ms` and `1s` are the same thing, the agent does duration parsing). It didn't feel like the right place to try and provide a full duration suffix spec tho, which is why I didn't specify. Do other agents also parse duration suffixes?

What do you think? Some options:
* Just change it to `SPLUNK_PROFILER_CALL_STACK_PERIOD_MS` and move on, skipping out on fancy (already built-in) duration parsing
* Specify that if units aren't provided it's assumed to be millis.

I'm thinking the latter.

iNikem: I like this convenience of java parsing, but probably we should follow the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/2fb3b181ae2ee928039cda3baa8acc1eea701ce8/specification/sdk-environment-variables.md#duration

Unless we can guarantee that all our language can provide reach parsing.

mateuszrzeszutek: > * Specify that if units aren't provided it's assumed to be millis.

^^ this 👍 
This is how all duration config properties already work in Java.

breedx-splk: I think that defaulting to ms is in keeping with the spec that @iNikem linked to. To me, the fact that java provides supplemental helpful ways of parsing duration doesn't seem to be in conflict with the spec.

pellared: It is not in conflict. See:
- https://github.com/open-telemetry/opentelemetry-specification/pull/2051#issuecomment-948957312
- https://github.com/open-telemetry/opentelemetry-specification/issues/1351

pellared: How about renaming the env var `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  so that it is similar to e.g. `OTEL_METRIC_EXPORT_INTERVAL`?

breedx-splk: > How about renaming the env var `SPLUNK_PROFILER_CALL_STACK_INTERVAL` so that it is similar to e.g. `OTEL_METRIC_EXPORT_INTERVAL`?

Well then that would break the original request to put the units in the env var name right?

pellared: I do not think the unit should be in the name.   OTel recommends ms as the default unit and allows support for setting the unit explicitly

breedx-splk: Sounds good to me!

specification/profiling.md
```diff
@@ -24,8 +24,10 @@ data.
 
 ## Call Stack Sampling
 
-An instrumentation library that has profiling capabilities MUST be able to sample call stacks
-at a fixed interval. This default interval MUST default to 1 second.
+An instrumentation library that has profiling capabilities MUST be able to sample 
+call stacks at a fixed interval. This interval MUST be user-configurable at startup 
+via the `SPLUNK_PROFILER_CALL_STACK_INTERVAL` environment variable. When not explicitly
+configured, the profiler SHALL use default period of 1 second.
```
iNikem, line 30:
```suggestion
configured, the profiler SHOULD use default period of 1 second.
```

pellared: I suggest merging https://github.com/signalfx/gdi-specification/pull/157 first and then also this PR could also add `SPLUNK_PROFILER_CALL_STACK_INTERVAL` to the table in `configuration.md`. Then the default value will be in the table.

breedx-splk:
Yeah, #157 dropped an expected bomb on this. Going to recreate.

PR #160, opened 2021-12-09
Add SPLUNK_PROFILER_CALL_STACK_INTERVAL

breedx-splk:
This is a remake of #150.

I thought it also made sense to make a default column, and the oddly centered content looked odd, so I just made it left aligned.

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -162,9 +162,9 @@ instance using the following environment variables:
 |----------------------------------------|---------|-----------------------------------------------------------------|
 | `SPLUNK_ACCESS_TOKEN`                  |         | Access token added to exported data. [1]                        |
 | `SPLUNK_METRICS_ENDPOINT`              |         | Endpoint for metrics data ingest.                               |
-| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10000   | Interval at which call stacks are sampled (in ms)               |
-| `SPLUNK_PROFILER_ENABLED`              | false   | Whether profiling is enabled. [2]                               |
-| `SPLUNK_PROFILER_LOGS_ENDPOINT`        | *       | Where profiling data is sent. Defaults to the value in `OTLP_EXPORTER_OTLP_ENDPOINT`  |
+| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10000   | Interval at which call stacks are sampled (in ms) [5]           |
+| `SPLUNK_PROFILER_ENABLED`              | false   | Whether profiling is enabled. [2] [5]                           |
+| `SPLUNK_PROFILER_LOGS_ENDPOINT`        | *       | Where profiling data is sent. Defaults to the value in `OTLP_EXPORTER_OTLP_ENDPOINT` [5]  |
 | `SPLUNK_REALM`                         | `none`  | Which realm to send exported data. [3]                          |
 | `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` |  true   | Whether `Server-Timing` header is added to HTTP responses. [4]  |
 
@@ -191,6 +191,8 @@ instance using the following environment variables:
   precedence.
 - [4]: If stitching of RUM spans and APM spans is desired then this parameter
   MUST be set to `true`.
+- [5]: Applies only to instrumentation libraries with profiling capabilities. 
+  Profiling is experimental status.
 
 In addition to Splunk-specific environment variables, the following
 [OpenTelemetry environment
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -162,7 +162,7 @@ instance using the following environment variables:
 |----------------------------------------|---------|-----------------------------------------------------------------|
 | `SPLUNK_ACCESS_TOKEN`                  | none    | Access token added to exported data. [1]                        |
 | `SPLUNK_METRICS_ENDPOINT`              | none    | Endpoint for metrics data ingest.                               |
-| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10s     | Interval at which call stacks are sampled                       |
+| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10s     | Interval at which call stacks are sampled.                      |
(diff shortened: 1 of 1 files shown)


specification/configuration.md
```diff
@@ -158,14 +158,15 @@ For all use-cases that support environment variables (e.g. applications and
...
-| `SPLUNK_PROFILER_ENABLED` (`false`)             | Whether profiling is enabled. [2]                                               |
-| `SPLUNK_PROFILER_LOGS_ENDPOINT` ()              | Where profiling data is sent. By default `OTLP_EXPORTER_OTLP_ENDPOINT` is used. |
-| `SPLUNK_REALM` (`none`)                         | Which realm to send exported data. [3]                                          |
-| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [4]                  |
+| Name                                   | Default | Description                                                     |
+|----------------------------------------|---------|-----------------------------------------------------------------|
+| `SPLUNK_ACCESS_TOKEN`                  | none    | Access token added to exported data. [1]                        |
+| `SPLUNK_METRICS_ENDPOINT`              | none    | Endpoint for metrics data ingest.                               |
+| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10s     | Interval at which call stacks are sampled                       |
```
flands, line 165: nit

```suggestion
| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10s     | Interval at which call stacks are sampled.                      |
```

specification/configuration.md
```diff
@@ -158,14 +158,15 @@ For all use-cases that support environment variables (e.g. applications and
...
+| `SPLUNK_PROFILER_ENABLED`              | false   | Whether profiling is enabled. [2]                               |
+| `SPLUNK_PROFILER_LOGS_ENDPOINT`        | *       | Where profiling data is sent. Defaults to the value in `OTLP_EXPORTER_OTLP_ENDPOINT`  |
+| `SPLUNK_REALM`                         | `none`  | Which realm to send exported data. [3]                          |
```
flands, line 168: What's the difference between a ticked none and a not ticked none? Perhaps blank default should not list anything in this column? Or I guess actual values can be ticked?

specification/configuration.md
breedx-splk: I never put periods at the end of partial sentence fragments. 🤓

specification/configuration.md
breedx-splk: I agree, this is confusing. I didn't actually know the answer and just brought it forward the way it was. Lemme go find the answer...

pellared: `none` is a valid and expected default value for the `SPLUNK_REALM`. It was discussed here: https://github.com/signalfx/gdi-specification/pull/148

The other values should be left blank

specification/configuration.md
```diff
@@ -158,14 +158,15 @@ For all use-cases that support environment variables (e.g. applications and
...
-| `SPLUNK_PROFILER_ENABLED` (`false`)             | Whether profiling is enabled. [2]                                               |
-| `SPLUNK_PROFILER_LOGS_ENDPOINT` ()              | Where profiling data is sent. By default `OTLP_EXPORTER_OTLP_ENDPOINT` is used. |
-| `SPLUNK_REALM` (`none`)                         | Which realm to send exported data. [3]                                          |
-| `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED` (`true`) | Whether `Server-Timing` header is added to HTTP responses. [4]                  |
+| Name                                   | Default | Description                                                     |
+|----------------------------------------|---------|-----------------------------------------------------------------|
+| `SPLUNK_ACCESS_TOKEN`                  | none    | Access token added to exported data. [1]                        |
+| `SPLUNK_METRICS_ENDPOINT`              | none    | Endpoint for metrics data ingest.                               |
+| `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10s     | Interval at which call stacks are sampled.                      |
```
pellared, line 165: I suggest changing it to `10000` as milliseconds is the only required (and default) duration unit. See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#duration

specification/configuration.md
breedx-splk: Thanks for the clarification, @pellared I'll get it changed.

specification/configuration.md
breedx-splk: Ok sure, we had a discussion in a previous PR about units, but I'm fine changing it to ms.

pellared: We can support units like `ms` but it is safer to use the the "required" unit in specs 😉

pellared (approved):

pellared:
I am just wondering if we can add new settings to "stable" Instrumentation Libraries configuration for a behavior that is "experimental" (Profiling)

flands:
> I am just wondering if we can add new settings to "stable" Instrumentation Libraries configuration for a behavior that is "experimental" (Profiling)

Oh, that is fun... @breedx-splk thoughts on this? Maybe we need a column to donate? Or are you comfortable marking stable?

breedx-splk:
> > I am just wondering if we can add new settings to "stable" Instrumentation Libraries configuration for a behavior that is "experimental" (Profiling)
> 
> Oh, that is fun... @breedx-splk thoughts on this? Maybe we need a column to donate? Or are you comfortable marking stable?

I think it's perfectly fine to add experimental features to stable components. Since `profiling.md` was assimilated I guess we did lose some tracking of what's experimental and what's stable. 

A column could help mitigate, but fear that the simple table (which I've already made more complicated by adding defaults) to turn into a convoluted stability matrix.

So far, there's really just one category of stuff that's experimental. Can we handle that with a footnote or something?

flands:
> So far, there's really just one category of stuff that's experimental. Can we handle that with a footnote or something?

Fine by me

kubawach (approved):
LGTM

kubawach:
> > So far, there's really just one category of stuff that's experimental. Can we handle that with a footnote or something?
> 
> Fine by me

@breedx-splk going to add the footnote? I would merge this ASAP :)

PR #163, opened 2022-01-25
serverless spec change

kubawach:
Serverless config spec change:
- defaults to OTLP over HTTP (as supported by direct ingest)
- don't require JT exporter (increases package size for eg Java)

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -277,7 +277,7 @@ Other requirements:
 **Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
 
 By default, serverless instrumentation libraries MUST send data directly
-to Splunk Observability Cloud (direct ingest). Therefore, following applies to exporter configuration:
+to Splunk Observability Cloud (direct ingest). Therefore, the following applies to exporter configuration:
 - `OTEL_TRACES_EXPORTER`
   - MUST default to `otlp` over HTTP, as currently supported by the ingest
   - MAY offer `jaeger-thrift-splunk` 
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -277,7 +277,11 @@ Other requirements:
 **Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
 
 By default, serverless instrumentation libraries MUST send data directly
-to Splunk Observability Cloud (direct ingest).
+to Splunk Observability Cloud (direct ingest). Therefore, following applies to exporter configuration:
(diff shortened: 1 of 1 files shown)


Kielek (approved):

owais (approved):

specification/configuration.md
```diff
@@ -277,7 +277,11 @@ Other requirements:
 **Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
 
 By default, serverless instrumentation libraries MUST send data directly
-to Splunk Observability Cloud (direct ingest).
+to Splunk Observability Cloud (direct ingest). Therefore, following applies to exporter configuration:
+- `OTEL_TRACES_EXPORTER`
+  - MUST default to `otlp` over HTTP, as currently supported by the ingest
+  - MAY offer `jaeger-thrift-splunk` 
+Serverless distribution MAY omit `jaeger-thrift-splunk` exporter due to package size optimisation.
```
owais, line 284: "may" could result in inconsistencies between different Lambda layers. We should consider Otel based Lambda always offering OTLP only. Is there a case where people deploying our new Otel based Lamba layers would want to send to Smart Agent?

kubawach: Node.js does not support otlp/grpc while our ingest works with otlp/http. Therefore there we have to use J/T.

flands: I think Owais point is different -- given we support OTLP HTTP why do we ever need jaeger-thrift-splunk? I would propose we drop the second line at a minimum. Perhaps just mention the MUST for OTLP

flands, line 284:
```suggestion
```

specification/configuration.md
```diff
@@ -277,7 +277,11 @@ Other requirements:
 **Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
 
 By default, serverless instrumentation libraries MUST send data directly
-to Splunk Observability Cloud (direct ingest).
+to Splunk Observability Cloud (direct ingest). Therefore, following applies to exporter configuration:
```
flands, line 280:
```suggestion
to Splunk Observability Cloud (direct ingest). Therefore, the following applies to exporter configuration:
```

specification/configuration.md
kubawach: Oh sure, it is a deviation from "master" instr lib spec so I wanted to say it here aloud (may instead of must).

flands (approved):

PR #166, opened 2022-03-25
profiling: pprof profile.proto format

laurit:
Describe sending profiling data in [pprof](https://github.com/google/pprof) [profile.proto](https://github.com/google/pprof/tree/master/proto) format. Pprof payload will contain a batch of samples. Attributes describing an individual sample, that in currently used `text` format are added to `LogRecord`, are added as label to pprof sample.
@vovencij @gsmirnov-splk please review

### DIFF
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -204,6 +204,6 @@ For each `allocation` sample:
 
 For each `cpu` sample:
 - label `source.event.period` of type `long` MUST contain the sampling period if this sample represents a periodic event
-- label `thread.status` of type `string` OPTIONALLY can be set to describe the state of the thread
+- label `thread.state` of type `string` OPTIONALLY can be set to describe the state of the thread
 
 Missing file name and function MUST be reported as `unknown`. Missing line number MUST be reported as `-1`.
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -186,7 +186,8 @@ after the `lineno`, separated by a space.
 
 #### PPROF Profile.proto Data Format
 
-`Profile.proto` is a data representation for profiling data defined at https://github.com/google/pprof/tree/master/proto.
+[`Profile.proto`](https://github.com/google/pprof/tree/master/proto) is a data representation for profiling data. It is
+independent of the type of data being collected and the sampling process used to collect that data.
 Log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
 contains either `allocation` or `cpu` samples determined by the data type specified for the log record.
 
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -190,21 +190,23 @@ after the `lineno`, separated by a space.
 independent of the type of data being collected and the sampling process used to collect that data.
 The log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
 contains either `allocation` or `cpu` samples determined by the data type specified for the log record.
+Data types `int64` and `string` are protocol buffer types, consult
+[protocol buffers documentation](https://developers.google.com/protocol-buffers/docs/proto#scalar).
 
 For each `allocation` and `cpu` sample:
 - label `source.event.name` of type `string` OPTIONALLY can contain the name of the event that triggered the sampling
-- label `source.event.time` of type `long` MUST be set to the time when the sample was taken
+- label `source.event.time` of type `int64` MUST be set to the unix time in millis when the sample was taken
 - label `trace_id` of type `string` MUST be set when sample was taken within a span scope
 - label `span_id` of type `string` MUST be set when sample was taken within a span scope
-- label `thread.id` of type `long` OPTIONALLY can be set to the thread identifier used by the runtime environment
+- label `thread.id` of type `int64` OPTIONALLY can be set to the thread identifier used by the runtime environment
 - label `thread.name` of type `string` OPTIONALLY can be set to the thread name used by runtime environment
-- label `thread.os.id` of type `long` MUST be set to the thread identifier used by the operating system
+- label `thread.os.id` of type `int64` MUST be set to the thread identifier used by the operating system
 
 For each `allocation` sample:
-- value of type `long` must be set to allocation size in bytes
+- value of type `int64` must be set to allocation size in bytes
 
 For each `cpu` sample:
-- label `source.event.period` of type `long` MUST contain the sampling period if this sample represents a periodic event
+- label `source.event.period` of type `int64` MUST contain the sampling period if this sample represents a periodic event
 - label `thread.state` of type `string` OPTIONALLY can be set to describe the state of the thread
 
 Missing file name and function MUST be reported as `unknown`. Missing line number MUST be reported as `-1`.
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -108,8 +108,8 @@ instances. For each `LogRecord` instance:
 - `source.event.period` MUST contain the sampling period if this `LogRecord` represents a periodic event
 - `source.event.name` OPTIONALLY can contain the name of the event that triggered the sampling
 - `memory.allocated` MUST contain the allocation size if this `LogRecord` represents a memory allocation event
-- `com.splunk.profiling.data.type` MUST be either `allocation` or `cpu`
-- `com.splunk.profiling.data.format` MUST be either `text` or `pprof-gzip-base64`
+- `profiling.data.type` MUST be either `allocation` or `cpu`
+- `profiling.data.format` MUST be either `text` or `pprof-gzip-base64`
 
 ### `LogRecord` Message Fields
 
@@ -188,7 +188,7 @@ after the `lineno`, separated by a space.
 
 [`Profile.proto`](https://github.com/google/pprof/tree/master/proto) is a data representation for profiling data. It is
 independent of the type of data being collected and the sampling process used to collect that data.
-Log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
+The log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
 contains either `allocation` or `cpu` samples determined by the data type specified for the log record.
 
 For each `allocation` and `cpu` sample:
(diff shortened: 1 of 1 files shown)


specification/semantic_conventions.md
```diff
@@ -183,3 +185,27 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
 - literal `)`
+
+#### PPROF Profile.proto Data Format
+
+`Profile.proto` is a data representation for profiling data defined at https://github.com/google/pprof/tree/master/proto.
+Log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
+contains either `allocation` or `profiling` samples determined by the data type specified for the log record.
+
+For each `allocation` and `profiling` sample:
+- label `source.event.period` of type `long` MUST contain the sampling period if this sample represents a periodic event
```
gsmirnov-splk, line 196: This is a bit confusing -- IIUC, `allocation` samples do not have this field, as it is only relevant for `profiling`. Also I wonder if we can put it in the value instead of an attribute?

gsmirnov-splk (approved):
All in all, lgtm!

specification/semantic_conventions.md
laurit: Moved it under `profiling` for now. We can make it a value instead of a label if you believe it is better that way.

gsmirnov-splk: Don't have strong feelings about that, but seems to be more consistent to put both as the "value". That's the approach we take on the backend too. @vovencij wdyt?

vsor-splk: My take on this is that interpreting `event.period` as value in the agent is not entirely semantically correct. One sample in pprof represents a measurement, which may have additional attributes besides the time the measurement was taken. Number of bytes allocated - this is a value of a measurement, no questions asked. In case of periodic thread call stack sampling, as cpu profiler does, the measurement is basically a count of 1 - the fact that at a particular time one particular call stack was observed for one time. Period at which we take samples - this is not a value, but an additional attribute. Yes, at the backend we do use this attribute as an approximate value when drawing a flamegraph, but this is not a value of that sample at the agent side.

specification/semantic_conventions.md
```diff
@@ -183,3 +185,27 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
 - ` <col>` or ` <col>:<col>` - OPTIONAL - If the runtime provides a column or column range, it MAY be provided
 after the `lineno`, separated by a space.
 - literal `)`
+
+#### PPROF Profile.proto Data Format
+
+`Profile.proto` is a data representation for profiling data defined at https://github.com/google/pprof/tree/master/proto.
```
vsor-splk, line 191: Should a link be formatted as a link?

mateuszrzeszutek (approved):

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
 - ` <col>` or ` <col>:<col>` - OPTIONAL - If the runtime provides a column or column range, it MAY be provided
 after the `lineno`, separated by a space.
 - literal `)`
+
+#### PPROF Profile.proto Data Format
+
+[`Profile.proto`](https://github.com/google/pprof/tree/master/proto) is a data representation for profiling data. It is
+independent of the type of data being collected and the sampling process used to collect that data.
+Log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
```
MrAlias, line 193:
```suggestion
The log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
```

specification/semantic_conventions.md
```diff
@@ -108,6 +108,8 @@ instances. For each `LogRecord` instance:
 - `source.event.period` MUST contain the sampling period if this `LogRecord` represents a periodic event
 - `source.event.name` OPTIONALLY can contain the name of the event that triggered the sampling
 - `memory.allocated` MUST contain the allocation size if this `LogRecord` represents a memory allocation event
+- `com.splunk.profiling.data.type` MUST be either `allocation` or `cpu`
+- `com.splunk.profiling.data.format` MUST be either `text` or `pprof-gzip-base64`
```
MrAlias, line 112: These names seem very Java-esq. Maybe instead:

```suggestion
- `profiling.data.type` MUST be either `allocation` or `cpu`
- `profiling.data.format` MUST be either `text` or `pprof-gzip-base64`
```

Also, these statements are normative only in value these fields must take on, but not whether these fields are to be included. Can These be updated to state if they should/must/may be included in payloads?

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+#### PPROF Profile.proto Data Format
+
+[`Profile.proto`](https://github.com/google/pprof/tree/master/proto) is a data representation for profiling data. It is
+independent of the type of data being collected and the sampling process used to collect that data.
+Log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
+contains either `allocation` or `cpu` samples determined by the data type specified for the log record.
+
+For each `allocation` and `cpu` sample:
+- label `source.event.name` of type `string` OPTIONALLY can contain the name of the event that triggered the sampling
+- label `source.event.time` of type `long` MUST be set to the time when the sample was taken
```
MrAlias, line 198: Can we specify more explicitly what a `long` type is. The underlying representation can very based on the language used. Maybe use `int64` or maybe explicitly call it a signed 64-bit (or 32-bit?) integer.

MrAlias, line 198: Please update this to specify this time is the unix time since epoc.

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+- label `trace_id` of type `string` MUST be set when sample was taken within a span scope
```
MrAlias, line 199: What encoding is used for this and the span ID strings?

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+- label `span_id` of type `string` MUST be set when sample was taken within a span scope
+- label `thread.id` of type `long` OPTIONALLY can be set to the thread identifier used by runtime environment
```
MrAlias, line 201:
```suggestion
- label `thread.id` of type `long` OPTIONALLY can be set to the thread identifier used by the runtime environment
```

Is this referring to the application space or just use space of the kernel?

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+- label `thread.name` of type `string` OPTIONALLY can be set to the thread name used by runtime environment
+- label `thread.native.id` of type `long` MUST be set to the thread identifier used by the operating system
```
MrAlias, line 203: Where does the term `native` come from? Maybe just use the term `os` directly?

```suggestion
- label `thread.os.id` of type `long` MUST be set to the thread identifier used by the operating system
```

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+
+For each `allocation` sample:
+- value of type `long` must be set to allocation size in bytes
```
MrAlias, line 206: How is this value expected to be stored in the log message. Is this meant to be the log message field?

mateuszrzeszutek: This is stored in the pprof message - which later is encapsulated into OTLP log message as its body

specification/semantic_conventions.md
pellared: I guess we should use the `proto` types

https://developers.google.com/protocol-buffers/docs/proto#scalar

specification/semantic_conventions.md
```diff
@@ -183,3 +185,28 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+independent of the type of data being collected and the sampling process used to collect that data.
+The log message will contain a gzip-compressed, base64-encoded protocol buffer conforming to `profile.proto`. Each message
+contains either `allocation` or `cpu` samples determined by the data type specified for the log record.
+
+For each `allocation` and `cpu` sample:
+- label `source.event.name` of type `string` OPTIONALLY can contain the name of the event that triggered the sampling
+- label `source.event.time` of type `long` MUST be set to the time when the sample was taken
+- label `trace_id` of type `string` MUST be set when sample was taken within a span scope
+- label `span_id` of type `string` MUST be set when sample was taken within a span scope
+- label `thread.id` of type `long` OPTIONALLY can be set to the thread identifier used by the runtime environment
```
MrAlias, line 201: Is this referring to the application space or just user space of the kernel?

specification/semantic_conventions.md
laurit: Changed `long` to `int64` and added link to protocol buffers doc. Alternatively could use `numeric` as the code generated from proto contains setters `setNum` and `setStr` it is only important to call the correct one of these.

laurit: changed `time` to `unix time in millis`

specification/semantic_conventions.md
laurit: it isn't important, protocol buffers will take care of the encoding (afaik it is utf-8)

specification/semantic_conventions.md
laurit: As @mateuszrzeszutek said it is defined in proto https://github.com/google/pprof/blob/b5a4dc8f4f2afdee77047b6aae3834140efc83ed/proto/profile.proto#L114

specification/semantic_conventions.md
laurit: the intention was that `thread.id` refers to applications specific thread id e.g. java has thread ids that aren't the same as os thread ids

specification/semantic_conventions.md
laurit: changed `MUST be either` to ` MUST be set to either`

specification/semantic_conventions.md
```diff
@@ -108,6 +108,8 @@ instances. For each `LogRecord` instance:
 - `source.event.period` MUST contain the sampling period if this `LogRecord` represents a periodic event
 - `source.event.name` OPTIONALLY can contain the name of the event that triggered the sampling
 - `memory.allocated` MUST contain the allocation size if this `LogRecord` represents a memory allocation event
+- `com.splunk.profiling.data.type` MUST be either `allocation` or `profiling`
+- `com.splunk.profiling.data.format` MUST be either `text` or `pprof-gzip-base64`
```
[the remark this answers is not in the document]
laurit: lets treat them as required in spec and let backend take care of the clients that use older version of spec where these aren't required

specification/semantic_conventions.md
```diff
@@ -183,3 +185,30 @@ provides a range of lines, the second `lineno` can be provided after a colon `:`
...
+[protocol buffers documentation](https://developers.google.com/protocol-buffers/docs/proto#scalar).
+
+For each `allocation` and `cpu` sample:
+- label `source.event.name` of type `string` OPTIONALLY can contain the name of the event that triggered the sampling
+- label `source.event.time` of type `int64` MUST be set to the unix time in millis when the sample was taken
+- label `trace_id` of type `string` MUST be set when sample was taken within a span scope
+- label `span_id` of type `string` MUST be set when sample was taken within a span scope
+- label `thread.id` of type `int64` OPTIONALLY can be set to the thread identifier used by the runtime environment
+- label `thread.name` of type `string` OPTIONALLY can be set to the thread name used by runtime environment
+- label `thread.os.id` of type `int64` MUST be set to the thread identifier used by the operating system
```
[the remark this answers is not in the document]
laurit: profiling team was ok with making it optional

MrAlias (approved):

MrAlias:
@laurit can you rebase to pull in the hyperlink fix? I think this is ready to merge when that is done.

PR #172, opened 2022-04-20
profiling: add attribute for truncated stack trace

laurit:
@vovencij @gsmirnov-splk please review
Add an attribute for marking truncated stack traces so that profiling backend would know that the stack trace has missing frames.


gsmirnov-splk:
The behavior when the stack trace was NOT truncated seems unspecified, is this on purpose? I guess the intent is to omit this attribute/label when it's not truncated. Right?

laurit:
> The behavior when the stack trace was NOT truncated seems unspecified, is this on purpose? I guess the intent is to omit this attribute/label when it's not truncated. Right?

yes, that was my intention.

gsmirnov-splk:
LGTM

mateuszrzeszutek (approved):

PR #176, opened 2022-07-28
Pprof related fixes

Kielek:
Based on https://github.com/signalfx/signalfx-dotnet-tracing/pull/565

Make the documentation more specific for the attributes needed for pprof/text format.

Changes should not affect existing implementation (like Java).

### DIFF
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -119,13 +119,13 @@ instances. For each `LogRecord` instance:
 
 ### `LogRecord` Message Fields
 
-- [Time](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp)
-  MUST be set to the time that the call stack was sampled.
 - [Body](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body)
   MUST be populated with appropriate payload for specified data type and format.
 
 #### `LogRecord` Message for `text` Data Format Specific Fields
 
+- [Time](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp)
+  MUST be set to the time that the call stack was sampled.
 - [TraceId](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-traceid)
   MUST be populated when a call stack has been sampled within a span scope.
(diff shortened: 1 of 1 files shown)


Kielek:
@vovencij. @seemk, @lachmatt, please check

lachmatt (approved):

pellared (approved):

mateuszrzeszutek (approved):
👍

specification/semantic_conventions.md
```diff
@@ -101,29 +101,35 @@ For each `InstrumentationLibraryLogs` instance:
...
 - `source.event.name` OPTIONALLY can contain the name of the event that triggered the sampling
 - `memory.allocated` MUST contain the allocation size if this `LogRecord` represents a memory allocation event
-- `profiling.data.type` MUST be set to either `allocation` or `cpu`
-- `profiling.data.format` MUST be set to either `text` or `pprof-gzip-base64`
 - `thread.stack.truncated` MUST be set to boolean `true` when this `LogRecord` does not contain the full stack trace
 
 ### `LogRecord` Message Fields
 
 - [Time](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp)
   MUST be set to the time that the call stack was sampled.
```
vsor-splk, line 123: I guess this should also move to the `text` payload?

Kielek: Fixed: dbf4517ce897dd50687ae15eb50ebcad49ef6d12

Kielek:
@mateuszrzeszutek, @seemk 
please check dbf4517ce897dd50687ae15eb50ebcad49ef6d12 - there is recommendation change. As I know, Java is sending this property on the wrong level.

mateuszrzeszutek:
> As I know, Java is sending this property on the wrong level.

No, we're sending this correctly: https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofCpuEventExporter.java#L92-L93

We do set the current time in the pprof `LogData` ([here](https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofLogDataExporter.java#L69)), but I believe that this is completely superfluous and probably could be removed without any harm.

seemk:
Node's serializer should be correct as well: https://github.com/signalfx/splunk-otel-js/blob/main/src/profiling/utils.ts#L107-L109.

seemk (approved):

MrAlias (approved):

PR #177, opened 2022-08-01
chore: clarify source.event.period unit

seemk:
The unit for `source.event.period` wasn't specified, however agents send it in milliseconds. But what if we ever want to use sub millisecond intervals? :eyes:

### DIFF
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -107,7 +107,7 @@ Inside each `InstrumentationLibraryLogs` instance is a list of `LogRecord`
 instances. For each `LogRecord` instance:
 
 - `com.splunk.sourcetype` MUST be set to the value `otel.profiling`
-- `source.event.period` MUST contain the sampling period if this `LogRecord` represents a periodic event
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

mateuszrzeszutek:
> But what if we ever want to use sub millisecond intervals?

I'm guessing we'll need a new attribute anyway, since the backend will need to interpret the data in a different way.

pjanotti (approved):
LGTM. cc/ @latchmatt @Kielek

breedx-splk:
> The unit for `source.event.period` wasn't specified, however agents send it in milliseconds. But what if we ever want to use sub millisecond intervals? 👀

At that point, I suppose we can roll back this change and add another field for units or something.

laurit (approved):

Kielek (approved):

lachmatt (approved):

Kielek (approved):

pjanotti (approved):

PR #186, opened 2022-09-01
MUST NOT use GitHub secrets

pellared:
I think it was the original intention.

@owais @flands PTAL

### DIFF
--- a/specification/repository.md
+++ b/specification/repository.md
@@ -54,6 +54,7 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
 
 - MUST NOT use [GitHub
   secrets](https://docs.github.com/en/actions/reference/encrypted-secrets)
+  - Exception: Secrets creates by GitHub with `GITHUB_` prefix.
 - MUST NOT use [Personal Access
   Tokens](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token)
 - MUST [limit permissions of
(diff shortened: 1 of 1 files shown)


specification/repository.md
```diff
@@ -52,10 +52,8 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
 
 ### GitHub Actions
 
-- MUST use [GitHub
-  secrets](https://docs.github.com/en/actions/reference/encrypted-secrets) to
-  store sensitive data (auth tokens, passwords) and limit their usage to only
-  required places
+- MUST NOT use [GitHub
+  secrets](https://docs.github.com/en/actions/reference/encrypted-secrets)
```
mateuszrzeszutek, line 56: Does `GITHUB_TOKEN` count as a secret?
If you run tests using GHCR you *have* to use that

pellared: Good comment.
For sure not (it is not explicitly set by us). I will add a remark that it does not count

pellared: PTAL https://github.com/signalfx/gdi-specification/pull/186/commits/97830548256438eec54a815cfba27a3f57bfaf86

mateuszrzeszutek (approved):

breedx-splk:
Noooo! Well, maybe this ship has already sailed, but I read this a different way. I read this to mean that "if you have a need for secrets in your repo, use GitHub secrets for these (instead of files or hard-coding or some other means)". I thought it was supporting/encouraging the use of GH secrets for sensitive data....but this change completely eliminates that....in fact now it's forbidden. 

I would like to see this change rolled back.

PR #192, opened 2022-10-31
Add tag protection requirement

pellared:
## Why

Per internal agreement.

Practice "defense in depth".

## What

Add a requirement to use https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules to increase the protection of our release train.

### DIFF
--- a/specification/repository.md
+++ b/specification/repository.md
@@ -128,6 +128,8 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
 - MUST have a signature or a checksum with signature for all release artifacts
   - SHOULD use Splunk signing key
 - MUST use [signed tags](https://docs.github.com/en/github/authenticating-to-github/signing-tags)
+- MUST have a [tag protection rule](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules)
+  for the release tags (e.g. `v*`)
 - MUST state version of OpenTelemetry repository built against if applicable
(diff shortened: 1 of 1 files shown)


Kielek (approved):

pellared:
> Seems like a reasonable idea to me. We might have to give a few Splunkers additional privs in some repos, however.

Maintainers are able to create a protected tag. Only admins are able to delete it. I think it is reasonable.

lachmatt (approved):

MrAlias (approved):

MrAlias:
@pellared please be sure to update splunk-otel-go

seemk:
Maintainers should be able to delete protected tags. Usually when you need tag deletion :hankey: has hit the fan and you don't want the extra overhead of catching an admin. Letting only admins delete tags does not provide more security, as the only critical part - **releasing** - happens when tags are created.

pellared:
> Maintainers should be able to delete protected tags. Usually when you need tag deletion 💩 has hit the fan and you don't want the extra overhead of catching an admin. Letting only admins delete tags does not provide more security, as the only critical part - **releasing** - happens when tags are created.

I disagree. The main benefit of such policy is that it reduces the changes to delete the tag after the release is published.

If everything is fine then you should never delete a tag 😉

pellared:
@seemk I would rather consider to grant `Admin` for `maintainers`  (on repositories) 😉 This will make e.g. setting up branch protection rules easier. Also `Admin` is requires to set up many things required by the GDI spec.

Issue #194, opened 2022-11-03
Deprecate Splunk-specific configs for metric


pellared:
We should use OTel metrics now.
Long term we don't want any Splunk specific config for metrics.

_Originally posted by @mateuszrzeszutek in https://github.com/signalfx/gdi-specification/issues/73#issuecomment-1301868422_

PR #196, opened 2022-11-10
Remove SPLUNK_METRICS_ENDPOINT from Instrumentation Libraries

pellared:
## Why

Fixes https://github.com/signalfx/gdi-specification/issues/194

The `SPLUNK_METRICS_ENDPOINT` is either not a stable setting or it is not supported in any of the existing instrumentation libraries.

1. In https://github.com/signalfx/splunk-otel-java it is marked as "experimental"
2. In https://github.com/signalfx/splunk-otel-js it looks like a undocumented functionality (so even less than "experimental")
3. https://github.com/signalfx/splunk-otel-go logs that it is not supported if it is set

Initially I wanted to add a "note" that this env var is not required, but a lean towards removing it from the spec. We should use/follow OTel metrics.

Also, during GDI spec review, none of the @signalfx/gdi-specification-maintainers was considering this env var as MUST HAVE.

At last, we do not want to implement it in https://github.com/signalfx/splunk-otel-dotnet

## What 

- In Instrumentation Libraries move `SPLUNK_METRICS_ENDPOINT` from "stable" to "experimental" (and optional) and also introduce other env vars defined here: https://github.com/signalfx/splunk-otel-java/blob/main/docs/advanced-config.md
- Fix issues reported by markdownlint

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -161,7 +161,6 @@ instance using the following environment variables:
 | Name                                   | Default | Description                                                                               |
 |----------------------------------------|---------|-------------------------------------------------------------------------------------------|
 | `SPLUNK_ACCESS_TOKEN`                  |         | Access token added to exported data. [1]                                                  |
-| `SPLUNK_METRICS_ENDPOINT`              |         | Endpoint for metrics data ingest.                                                         |
 | `SPLUNK_PROFILER_CALL_STACK_INTERVAL`  | 10000   | Interval at which call stacks are sampled (in ms) [5]                                     |
 | `SPLUNK_PROFILER_ENABLED`              | false   | Whether CPU profiling is enabled. [2] [5]                                                 |
 | `SPLUNK_PROFILER_LOGS_ENDPOINT`        | *       | Where profiling data is sent. Defaults to the value in `OTLP_EXPORTER_OTLP_ENDPOINT` [5]  |


pjanotti (approved):
Per @breedx-splk comment, it makes sense to list it somewhere else as experimental, but, not on the "stable" table. LGTM.

pellared:
> There are 14 code uses in the signalfx org: https://github.com/search?q=org%3Asignalfx+SPLUNK_METRICS_ENDPOINT&type=code

I checked all of them. Take notice that some are related to "serverless"

> Java calls many of its configuration items experimental to set a clear expectation that the configuration names could change.
Java uses even env vars which are not 

Also see: https://github.com/signalfx/gdi-specification/blob/c9216b2d05faa82da4f5a62d801476994f56ff56/specification/configuration.md#L12-L19

I am not sure if the fact that Java has it is a reason to have it in GDI spec. Especially that Java has env vars that are NOT mentioned in the GDI spec (therefore it violates the spec).

Moreover, I think that right now I think that we SHOULD NOT implement `SPLUNK_METRICS_ENDPOINT` which is not inline with the current recommendation that says that experimental (e.g. profiling) SHOULD be supported.

This is the reason why I proposed to remove this from the spec.

Maybe the consensus would be to put it as "Deprecated". I could pull also other Java metric-related env vars so that it does not violate the GDI spec 😉 

@breedx-splk What do you propose? I do not have any strong opinion.

breedx-splk:
> Especially that Java has env vars that are NOT mentioned in the GDI spec (therefore it violates the spec).

This is blatantly false. The spec is not intended to be comprehensive, and languages/distributions/instrumentations are allowed to have supplemental functionality and configuration.

> @breedx-splk What do you propose? I do not have any strong opinion. Do you think a "Deprecated" table would be the best way going forward?

I hadn't really intended to propose anything. We can choose to remove it, but Java will still need a way for users to specify where to send metrics.

breedx-splk:
> The spec says something else 🤷

Wow. Do you think that is supposed to also include upstream configuration items? Cuz if so, I have about 99 issues to create....better get busy!

pellared:
@breedx-splk PTAL: https://github.com/signalfx/gdi-specification/pull/196/commits/341a14e819b0fdb5d235efa770f4c0accda104fd

Upstream configurations are not needed. This is only about Splunk-sepcific configuration 😉

specification/configuration.md
```diff
@@ -227,12 +228,23 @@ are required.
...
+    ```log
     jaeger-thrift-splunk trace exporter is deprecated and may be removed in a future major release. Use the default 
     OTLP exporter instead, or set the SPLUNK_REALM and SPLUNK_ACCESS_TOKEN environment variables to send 
     telemetry directly to Splunk Observability Cloud.
     ```
 
+
+The following environment variables are NOT required
+(the instrumentation libraries MAY support them):
```
pjanotti, line 240:
```suggestion
(instrumentation libraries MAY support them):
```

specification/configuration.md
```diff
@@ -288,31 +301,33 @@ Other requirements:
...
+  If `SPLUNK_REALM` is set, `SPLUNK_ACCESS_TOKEN` MUST be set as well.
+
+  With `SPLUNK_REALM` set, both traces and metrics exporter endpoints will have following values:
 
-    With `SPLUNK_REALM` set, both traces and metrics exporter endpoints will have following values:
-    - traces (in case of `otlp`): `https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp`
-    - traces (all other cases):`https://ingest.REALM.signalfx.com/v2/trace`
-    - metrics: `https://ingest.${SPLUNK_REALM}.signalfx.com`
+  - traces (in case of `otlp`): `https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp`
+  - traces (all other cases):`https://ingest.REALM.signalfx.com/v2/trace`
```
pjanotti, line 324:
```suggestion
  - traces (all other cases):`https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace`
```

specification/configuration.md
pjanotti: @breedx-splk are these "workaround" settings or something that eventually should reach the GDI spec?

pjanotti (approved):
The broad agreement is about removing the mentioned setting from stable. There is a big issue with the scope of GDI spec. Let's handle that as a separate issue.

pjanotti:
@pellared @breedx-splk my first take is contrary to what is suggested at https://github.com/signalfx/gdi-specification/blob/c9216b2d05faa82da4f5a62d801476994f56ff56/specification/configuration.md?plain=1#L15-L19

The GDI spec should cover the minimal env vars required on all stable instrumentations. It seems unattainable to do otherwise. Anyway, it would be better to handle it separately from this review.

specification/configuration.md
pjanotti: I'm having second thoughts if we want to document such config options here. Unless this is something that we think that eventually, all instrumentation should have, it will be easier to just remove them.

seemk:
I agree that we should prefer OTel specific env vars, so for repositories that have moved to OTel metrics, this env var is not really necessary any longer.

In the JS distro the preference is `programmatically -> SPLUNK_METRICS_ENDPOINT -> (handled by OTel) OTEL_EXPORTER_OTLP_METRICS_ENDPOINT -> OTEL_EXPORTER_OTLP_ENDPOINT`.

`SPLUNK_METRICS_ENDPOINT` is a remnant of the SignalFx metrics SDK, not really necessary any longer so it could be deprecated as long as the SDK supports OTel fully?

The preferred endpoint setup should anyway (as long as OTel is supported for all signals) be:

* For direct to cloud:
```
SPLUNK_REALM=us0
SPLUNK_ACCESS_TOKEN=...
launch
```
* To collector:
```
OTEL_EXPORTER_OTLP_ENDPOINT=collector:4317
launch
```

pellared:
I see 3 options (from the most to least aggressive):

1. remove this configuration from the spec (it was not said to be "stable" in any distro) - I think it was a bug that this env var ended up in the "stable" section - _this was the initial version of the PR_
2. deprecate the env var
3. move the env var to "optional experimental" configuration - _current PR version, as (based on existing comments) I think it is the easiest way to go forward_

seemk:
I'm up for removing it from the spec and leave it to distro authors not to break user code based on whatever versioning rules for a given platform :stuck_out_tongue:

pellared:
> > Especially that Java has env vars that are NOT mentioned in the GDI spec (therefore it violates the spec).
> 
> This is blatantly false. The spec is not intended to be comprehensive, and languages/distributions/instrumentations are allowed to have supplemental functionality and configuration.
> 
> > @breedx-splk What do you propose? I do not have any strong opinion. Do you think a "Deprecated" table would be the best way going forward?
> 
> I hadn't really intended to propose anything. We can choose to remove it, but Java will still need a way for users to specify where to send metrics.

@breedx-splk  I created https://github.com/signalfx/gdi-specification/issues/198

mateuszrzeszutek:
> My preference is **1 (removal)**. However, I we also start with **3 (optional experimental)** if anyone would have objections.

I think we should remove these config options - even if some instrumentation libraries (java) still use them. `SPLUNK_METRICS_*` configs and the entire Java metrics implementation are just tech debt, I believe we should not be codifying this crap in any spec. Long term, in the (hopefully close) future we're going to be using OTel metrics everywhere.

mateuszrzeszutek:
> The GDI spec should cover the minimal env vars required on all stable instrumentations. It seems unattainable to do otherwise.

Even OTel spec does that -- language SDKs/instrumentations have a ton of experimental configs that are not specced. Some of is is obviously going to change in the future (once we get some kind of general "instrumentation configuration" spec), but I think that we'll always have language/implementation specific settings.

mateuszrzeszutek (approved):

pellared:
Based on feedback I decided to revert the PR to the original state: **removal**).

By @breedx-splk 

> We can choose to remove it, but Java will still need a way for users to specify where to send metrics.

By @seemk 

> I'm up for removing it from the spec and leave it to distro authors not to break user code based on whatever versioning rules for a given platform

By @mateuszrzeszutek 

> I think we should remove these config options - even if some instrumentation libraries (java) still use them. 

I also created https://github.com/signalfx/gdi-specification/issues/198 to address the issue of introducing language/implementation specific settings.

specification/configuration.md
pellared: Per @mateuszrzeszutek comment: we do not want to have these settings. I reverted the PR to the original state.

seemk (approved):

pjanotti (approved):

pellared:
@breedx-splk PTAL

breedx-splk (approved):
As it reads now (one line removal of the MUST for `SPLUNK_METRICS_ENDPOINT`) I think I'm good with this. The way I read this is that implementations will no longer be required to implement `SPLUNK_METRICS_ENDPOINT`, but to me that doesn't preclude implementations from having it (and yeah #198 helps with this).

Sorry if that was the original intention and I missed the point. I just wanted to make sure that removal of this didn't instantly make java "in violation". 

We've got other approvals as well, so let's do this and continue forward.

theletterf:
@breedx-splk @seemk LMK if this will have an impact on our current config / get started docs. Currently, we're using `SPLUNK_METRICS_ENDPOINT` for Java and NodeJS.

PR #199, opened 2022-11-14
Update OTLP exporter default requirements

pellared:
## Why

The OTel specification has been updated some time ago: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol

Now, `http/protobuf` is the recommended default transport protocol. 

Also for technical reasons, in .NET our distro, we do not want to default to gRPC protocol. More [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#otlp)

Our backend accepts OTLP over `http/protobuf` on: `https://ingest/.{REALM}.signalfx.com/v2/trace/otlp`

## What

1. Allow defaulting to `http/protobuf`.
2. Remove the default value as this is the same on as [specified in Otel](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)

This change is backwards compatible.

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -222,7 +222,7 @@ are required.
 - Zipkin exporter
   - Distribution MUST NOT list Zipkin exporter as supported (not supported by Smart Agent)
 - `OTEL_TRACES_EXPORTER`
-  - Non-RUM distribution MUST default to `otlp` over gRPC with an endpoint of `localhost:4317`
+  - Non-RUM distribution MUST default to `otlp` using `grpc` or `http/protobuf` transport protocol.
   - Non-RUM distribution MAY offer `jaeger-thrift-splunk` that defaults to `http://127.0.0.1:9080/v1/trace`.
     **NOTE: `jaeger-thrift-splunk` is deprecated.**
     If the user selects `jaeger-thrift-splunk`, distributions MUST log a deprecation warning and suggest an alternate method. For example:


mateuszrzeszutek:
> I do not remember if our backend supports OTLP over gRPC for spans

It does, that's what we're using in the Java distro.

I vaguely remember there were some problems with gRPC in some languages (Ruby?) in the OTel spec, which is why the default got switched from `grpc` to `http/protobuf`. OTel Java was already stable at that point, which is why we're using `grpc` by default.

mateuszrzeszutek (approved):

pellared:
> It does, that's what we're using in the Java distro.

I meant ``http/protobuf` (not gRPC)🤦

pellared:
Our backend accepts OTLP over `http/protobuf` on: `https://ingest/.{REALM}.signalfx.com/v2/trace/otlp`.

Thanks @seemk ❤️

pjanotti (approved):

specification/configuration.md
```diff
@@ -222,7 +222,7 @@ are required.
 - Zipkin exporter
   - Distribution MUST NOT list Zipkin exporter as supported (not supported by Smart Agent)
 - `OTEL_TRACES_EXPORTER`
-  - Non-RUM distribution MUST default to `otlp` over gRPC with an endpoint of `localhost:4317`
+  - Non-RUM distribution MUST default to `otlp` using `grpc` or `http/protobuf` transport protocol.
```
breedx-splk, line 225: This loses the default endpoint specification. In other words, it used to be clear from reading this what wire format and protocol should be used, but also where it should be sent. Now it just says what format and protocols might be used, but not where to send it.

Can we get default endpoints specified?

pellared: > Remove the default value as this is the same on as [specified in Otel](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)

breedx-splk: Makes sense, thanks for clarifying. It's redundant to add it here. 👍🏻

pellared:
@breedx-splk Is there anything else I should address? Or you have just forgotten to approve? 😉

Kielek (approved):

breedx-splk (approved):
Looks good to me. Thanks for the clarifications.

Issue #63, opened 2021-04-30
When new minor or major tag is pushed automatically open issue in associated projects


flands:
**Is your feature request related to a problem? Please describe.**
The GDI specification will cut new releases in the future and GDI project will need to update. This needs to be automated so GDI project maintainers are aware of changes.

**Describe the solution you'd like**
When a minor or major tag is pushed all GDI project following the spec automatically get an issue opened asking to address

**Describe alternatives you've considered**
GDI spec maintainers manually creating these tickets

**Additional context**
This proposal does require keeping state of which GDI projects need to be notified

pellared:
Doing it automatically via GitHub Action would require having a bot account with PAT.

For security reasons, I think it would be better if we prepare a script that could be invoked by GDI spec maintainer as part of the release process.

PR #201, opened 2022-11-14
Add a list of GDI repositories

pellared:
## Why

Towards https://github.com/signalfx/gdi-specification/issues/63

## What

Add a list of GDI repositories to be clear for what repositories the spec should be applied. 

It should also facilitate the releasing process. We could create issues in the GDI repos when a new version of the spec is released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 ![Release](https://img.shields.io/github/v/tag/signalfx/gdi-specification?include_prereleases&style=for-the-badge)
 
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (approved):

seemk (approved):
Do note that `splunk-otel-cpp` is unmaintained

pjanotti (approved):

breedx-splk (approved):
LGTM

PR #207, opened 2023-01-13
Drop Jira app requirement

Kielek:
Introduced in #50 .

I do not see any benefits as we do not use any JIRA-GitHub integration.


Kielek:
@emaderer, @flands what do you think?

mateuszrzeszutek (approved):

mateuszrzeszutek:
I don't think we're actually using this plugin anywhere, even if it is enabled.

pellared (approved):

PR #211, opened 2023-03-02
Add a recommendation to set process resource attributes

pellared:
## Why

Maintenance 😉  These resource attributes are already set by our distributions.

## What 

Add a recommendation to set process resource attributes.

### DIFF
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -26,7 +26,7 @@ All Splunk distributions of OpenTelemetry,
 - SHOULD set the following resource attributes when applicable:
   - `telemetry.auto.version`
 
-- SHOULD set [process an process runtime resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.18.0/specification/resource/semantic_conventions/process.md)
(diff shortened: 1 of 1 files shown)


specification/semantic_conventions.md
```diff
@@ -26,6 +26,8 @@ All Splunk distributions of OpenTelemetry,
 - SHOULD set the following resource attributes when applicable:
   - `telemetry.auto.version`
 
+- SHOULD set [process an process runtime resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.18.0/specification/resource/semantic_conventions/process.md)
```
pellared, line 29: note: The OTel spec is "experimental". However, our distros are already setting these attributes. That is why I put a `SHOULD`. Still I am unsure if this should land in "stable" section or we should it be under "experimental".

pellared: Now I see that even https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk is marked as `Experimental`...

theletterf: `process an process` (is that right?)

pellared: typo fixed 😆 it should be `and`.

breedx-splk (approved):

PR #212, opened 2023-03-03
Add a recommendation to collect Runtime Environment Metrics

pellared:
## Why

Maintenance 😉 AFAIK the runtime metrics instrumentation is already in all our distributions.

## What 

Add a recommendation to set process resource attributes.

I do not want to make it as `MUST` because the lack of it should not block a distro release. 

I do not want to to make it as `Stable` because it is not stable in the OTel spec.


Kielek (approved):

pjanotti (approved):

mateuszrzeszutek (approved):
I'm okay with merging that - the Java distro probably won't fulfill these criteria for a while (for various reasons), but I think these are perfectly sound recommendations to any instrumentation agent/library that's not burdened with historical experimental stuff.

PR #218, opened 2023-03-24
Relax the meaning of allocation size

laurit:
For allocation profiling we wish to use [ObjectAllocationSample](https://bestsolution-at.github.io/jfr-doc/openjdk-17.html#jdk.ObjectAllocationSample) event that does not provide a allocation size in bytes
@gsmirnov-splk please review

### DIFF
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -226,7 +226,7 @@ For each `allocation` and `cpu` sample:
 
 For each `allocation` sample:
 
-- value of type `int64` must be set to allocation size in bytes
+- value of type `int64` must be set to a value that allows for comparing relative weights of the allocations, such as allocation size in bytes
 
 For each `cpu` sample:
 


pellared (approved):

gsmirnov-splk:
LGTM, thanks

mateuszrzeszutek (approved):

specification/semantic_conventions.md
```diff
@@ -226,7 +226,7 @@ For each `allocation` and `cpu` sample:
 
 For each `allocation` sample:
 
-- value of type `int64` must be set to allocation size in bytes
+- value of type `int64` must be set to a value that allows for comparing relative weights of the allocations, such as allocation size in bytes
```
Kielek, line 229: How it should be comparable? Within one process execution? Globally? Across all supported languages?

laurit: Within a single application. As far as I understand these values are aggregated and used to visualise the allocation pressure from a given sample. The current understanding is that the `weight` as described in https://bestsolution-at.github.io/jfr-doc/openjdk-17.html#jdk.ObjectAllocationSample is suitable for this purpose.

Kielek (approved):

seemk (approved):

PR #225, opened 2023-03-30
Release lifecycle marking

akubik-splunk:
Initial proposal for rules for marking status of released version. It will allow to clearly define window of support (all releases currently supported)  and pass the information to the end users. 
If the change is accepted we all released versions described in the changelog should be updated with appropriate status

### DIFF
--- a/specification/templates/CHANGELOG.md
+++ b/specification/templates/CHANGELOG.md
@@ -17,6 +17,15 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Version 0.1.0 - 2021-04-01
 
+### Life cycle status 
+
+Use one of:	
+-	*Latest*
+-	*Active*
+-	*Deprecated:* End of Support expected on 2024-04-01
+-	*End of Support:* Support has ended on 2024-04-01
+
+
 ### General
 
 - This release marks the first release of the repository.
--- a/specification/templates/CHANGELOG.md
+++ b/specification/templates/CHANGELOG.md
@@ -20,8 +20,7 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 ### Life cycle status 
 
 Use one of:	
--	*Latest*
--	*Active*
+-   *Supported*
 -	*Deprecated:* End of Support expected on 2024-04-01
 -	*End of Support:* Support has ended on 2024-04-01
 
--- a/specification/templates/CHANGELOG.md
+++ b/specification/templates/CHANGELOG.md
@@ -17,13 +17,7 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Version 0.1.0 - 2021-04-01
 
-### Life cycle status
-
-Use one of:
-
-- *Supported*
-- *Deprecated:* End of Support expected on 2024-04-01
-- *End of Support:* Support has ended on 2024-04-01
+Status: **Deprecated** (End of Support expected on 2024-04-01)
 
 ### General
 
--- a/specification/templates/CHANGELOG.md
+++ b/specification/templates/CHANGELOG.md
@@ -17,13 +17,13 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Version 0.1.0 - 2021-04-01
 
-### Life cycle status 
+### Life cycle status
 
-Use one of:	
--   *Supported*
--	*Deprecated:* End of Support expected on 2024-04-01
--	*End of Support:* Support has ended on 2024-04-01
+Use one of:
 
+- *Supported*
+- *Deprecated:* End of Support expected on 2024-04-01
+- *End of Support:* Support has ended on 2024-04-01
 
 ### General
 
--- a/specification/versioning.md
+++ b/specification/versioning.md
@@ -86,25 +86,26 @@ delivered to the users. If the release is also distributed via other
 distribution channels (e.g. CDN, Registries, AppStores) it `MUST` state
 its life cycle status according to following rules by any means available
 in the given software distribution solution.
- 
-Each release `MUST` clearly state the life cycle 
-status and described as: Supported, Deprecated, End of Support. 
-At minimum the required information should be included in [CHAANGELOG.md](https://github.com/signalfx/gdi-specification/blob/main/specification/templates/CHANGELOG.md)
-and equivalent mechanism provided by system used for distributing software. 
 
-If new version is released previously released versions `MUST` be evaluated 
+Each release `MUST` clearly state the life cycle
+status and described as: Supported, Deprecated, End of Support.
+At minimum the required information should be included in [CHAANGELOG.md](https://github.com/signalfx/gdi-specification/blob/main/specification/templates/CHANGELOG.md)
+and equivalent mechanism provided by system used for distributing software.
+
+If new version is released previously released versions `MUST` be evaluated
 according to stability guarantees and marked as:
+
 - Latest release `SHOULD` be marked as **Supported** indicating that this is
   version recommended to new and existing users
-- If applicable latest release of previous `MAJOR` line that is still under 
-  active development `MUST` be marked as **Supported** indicating that previous 
-  `MAJOR` line is under active development and users not ready to adopt 
+- If applicable latest release of previous `MAJOR` line that is still under
+  active development `MUST` be marked as **Supported** indicating that previous
+  `MAJOR` line is under active development and users not ready to adopt
   latest `MAJOR` version may use previous versions.
-- Release which was superseded by newer version, thus has entered deprecation 
-  period `MUST` be marked as **Deprecated** and `MUST` indicated planned 
-  end of support date. This indicates that users should plan for upgrade to 
+- Release which was superseded by newer version, thus has entered deprecation
+  period `MUST` be marked as **Deprecated** and `MUST` indicated planned
+  end of support date. This indicates that users should plan for upgrade to
   one of supported versions before date given.
-- Release which deprecation period has ended `MUST` be marked as 
-  **End of support** and `MUST` indicate date on which the release was marked 
-  as end of support. This indicates that support for given release is no longer 
+- Release which deprecation period has ended `MUST` be marked as
+  **End of support** and `MUST` indicate date on which the release was marked
+  as end of support. This indicates that support for given release is no longer
   provided and the users should immediately upgrade to newer supported version.
--- a/specification/versioning.md
+++ b/specification/versioning.md
@@ -80,11 +80,12 @@ until the end of that component’s existence:
 
 ## Release life cycle management
 
-Each repository `MUST` use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) to create snapshot of components
-and features delivered to the users. If release is also distributed via other 
-distribution channels (e.g. CDN, Registries, AppStores) it `MUST` be described 
-according to following rules by any means available in the given software 
-distribution solution. 
+Each repository `MUST` use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases)
+to create a versioned snapshot of components and features
+delivered to the users. If the release is also distributed via other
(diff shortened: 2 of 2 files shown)


specification/versioning.md
```diff
@@ -77,3 +77,33 @@ until the end of that component’s existence:
...
+
+Each repository `MUST` use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) to create snapshot of components
+and features delivered to the users. If release is also distributed via other 
+distribution channels (e.g. CDN, Registries, AppStores) it `MUST` be described 
+according to following rules by any means available in the given software 
+distribution solution. 
+ 
+Each release `MUST` clearly state the life cycle 
+status and described as: Active, Latest, Deprecated, End of Support. 
+At minimum the required information should be included in [CHAANGELOG.md](https://github.com/signalfx/gdi-specification/blob/main/specification/templates/CHANGELOG.md)
```
pellared, line 91:
```suggestion
At minimum the required information should be included in 
[CHANGELOG.md](https://github.com/signalfx/gdi-specification/blob/main/specification/templates/CHANGELOG.md)
```

specification/versioning.md
```diff
@@ -77,3 +77,33 @@ until the end of that component’s existence:
...
+and equivalent mechanism provided by system used for distributing software. 
+
+If new version is released previously released versions `MUST` be evaluated 
+according to stability guarantees and marked as:
+- Latest release `SHOULD` be marked as **Latest** indicating that this is
+  version recommended to new and existing users
```
pellared, line 97: Should this not be only applied for "stable" releases. E.g. beta/rc releases are usually not marked as latest.

specification/versioning.md
```diff
@@ -77,3 +77,33 @@ until the end of that component’s existence:
...
+  `MAJOR` line is under active development and users not ready to adopt 
+  latest `MAJOR` version may use previous versions.
+- Release which was superseded by newer version, thus has entered deprecation 
+  period `MUST` be marked as **Deprecated** and `MUST` indicated planned 
+  end of support date. This indicates that users should plan for upgrade to 
+  latest or active version respectively before date given.
+- Release which deprecation period has ended `MUST` be marked as 
+  **End of support** and `MUST` indicate date on which the release was marked 
+  as end of support. This indicates that support for given release is no longer 
+  provided and the users should immediately upgrade to newer version.
```
pellared, line 109: Can we remove "releases" which are Out of Support? E.g. some package management systems offers retracting packages.

pellared (commented):
Personally, I would prefer to have one definition of life cycle status which we could use on all levels:

- releases of components
- features of components
- GDI specification
- other? 😅 

The other think is how we "mark" the life cycle status of a release.

If we have a consensus on it then I propose to make two separate issues. The first one, to refine the life cycle status definition. And a second, to define how releases are affected by its life cycle statuses.

specification/versioning.md
```diff
@@ -77,3 +77,33 @@ until the end of that component’s existence:
...
+- Latest release `SHOULD` be marked as **Latest** indicating that this is
+  version recommended to new and existing users
+- If applicable latest release of previous `MAJOR` line that is still under 
+  active development `MUST` be marked as **Active** indicating that previous 
+  `MAJOR` line is under active development and users not ready to adopt 
+  latest `MAJOR` version may use previous versions.
+- Release which was superseded by newer version, thus has entered deprecation 
+  period `MUST` be marked as **Deprecated** and `MUST` indicated planned 
+  end of support date. This indicates that users should plan for upgrade to 
+  latest or active version respectively before date given.
```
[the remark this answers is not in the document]
akubik-splunk: Minor and Major:
1. Major version can be used and live for more than 12M
2. In some cases Major releases may be rare

specification/versioning.md
```diff
@@ -77,3 +77,33 @@ until the end of that component’s existence:
...
+## Release life cycle management
+
+Each repository `MUST` use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) to create snapshot of components
+and features delivered to the users. If release is also distributed via other 
+distribution channels (e.g. CDN, Registries, AppStores) it `MUST` be described 
+according to following rules by any means available in the given software 
+distribution solution. 
+ 
+Each release `MUST` clearly state the life cycle 
+status and described as: Active, Latest, Deprecated, End of Support.
```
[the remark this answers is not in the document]
akubik-splunk: It is possible to use the same list of labels/statuses on levels however the meaning might be little different depending on the context also the application might be different so to keep it clear we would have to map each status with a description on what it actually means in each context (release, component, spec etc) The final list would look like this:
(None) 
Experimental
Stable (same as active)
Deprecated 
End of Support 
Removed 

Additionally it should be possible to annotate Deprecated and End of Support with a date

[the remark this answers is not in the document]
pellared: Resolving per https://github.com/signalfx/gdi-specification/pull/225#discussion_r1153220625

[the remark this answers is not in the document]
akubik-splunk: The main issue is the context of in which "Stable" and "Experimental" is used - it work well for functional maturity it works less for lifecycle e.g. "this release has stable support", "this release has experimental support" or is "supported as experimental" thus was the proposal to keep a parallel set of labels. As stated in other comment we can make use one set but we need to explain what they mean depending on which context we use them in

specification/versioning.md
akubik-splunk: We could once the EoS is reached we should probably have some short period after which we could bury a the release and l leave only tombstone behind or get rid of it all together

specification/versioning.md
akubik-splunk: Good example - we should keep the latest at the stable level as we do not want to promote latest on experimental releases. The new release  release should be marked as experimental or experimental, supported 
Btw no I think about it - instead of "Active" we should probably use "Supported"

akubik-splunk: So actually instead of 
-	*Latest*
-	*Active*
-	*Deprecated:* End of Support expected on 2024-04-01
-	*End of Support:* Support has ended on 2024-04-01

We could have 
-       "Supported"
-	*Deprecated:* End of Support expected on 2024-04-01
-	*End of Support:* Support has ended on 2024-04-01

specification/templates/CHANGELOG.md
```diff
@@ -17,6 +17,15 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
...
 
+### Life cycle status 
+
+Use one of:	
+-	*Latest*
+-	*Active*
+-	*Deprecated:* End of Support expected on 2024-04-01
+-	*End of Support:* Support has ended on 2024-04-01
+
+
```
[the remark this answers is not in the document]
akubik-splunk: @pellared I want to limit the current PR to pint No. 3 - We can add the component level clarifications separately. Please let me know if you are ok with this

specification/versioning.md
akubik-splunk: @pellared change introduced and committed to PR. Please take a look

specification/versioning.md
```diff
@@ -77,3 +77,35 @@ until the end of that component’s existence:
...
+  `MAJOR` line is under active development and users not ready to adopt
+  latest `MAJOR` version may use previous versions.
+- Release which was superseded by newer version, thus has entered deprecation
+  period `MUST` be marked as **Deprecated** and `MUST` indicated planned
+  end of support date. This indicates that users should plan for upgrade to
+  one of supported versions before date given.
+- Release which deprecation period has ended `MUST` be marked as
+  **End of support** and `MUST` indicate date on which the release was marked
+  as end of support. This indicates that support for given release is no longer
+  provided and the users should immediately upgrade to newer supported version.
```
pellared, line 111: I made some nit changes, remove some IMO unnecessary sentences or wording, I tried to restructure it a little to make it easier to read and validate when adopting the spec.

```suggestion
Each release MUST clearly state the life cycle
status and described as one of:
**Supported**, **Deprecated**, **End of Support**.
At minimum the required information is included in [CHANGELOG.md](templates/CHANGELOG.md)
and equivalent mechanism provided by system used for distributing software.

Each repository MUST use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases)
to create a versioned snapshot of components
delivered to the users.

The release the life cycle status MUST be included
in [CHANGELOG.md](templates/CHANGELOG.md).

When the component is also distributed via other channels
(e.g. CDN, Registries, AppStores) it MUST state
its life cycle status according to following rules by any means available
in the given software distribution solution.

When a new version is released,
all previously released versions MUST be evaluated
according to stability guarantees and marked as:

- If applicable latest release of previous `MAJOR` line that is still under
  active development MUST remain as **Supported** indicating that previous
  `MAJOR` line is under active development and users not ready to adopt
  the latest `MAJOR` version may use previous versions.

- Release which was superseded by newer version, thus has entered deprecation
  period `MUST` be marked as **Deprecated** and MUST indicated planned
  end of support date. This indicates that users should plan for upgrade to
  one of supported versions before date given.

- Release which deprecation period has ended MUST be marked as
  **End of support** and MUST indicate date on which the release was marked
  as end of support. This indicates that support for given release is no longer
  provided and the users should immediately upgrade to newer supported version.
```

specification/versioning.md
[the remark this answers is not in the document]
pellared: I like this idea 👍

specification/versioning.md
```diff
@@ -77,3 +77,35 @@ until the end of that component’s existence:
...
+
+Each repository `MUST` use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases)
+to create a versioned snapshot of components and features
+delivered to the users. If the release is also distributed via other
+distribution channels (e.g. CDN, Registries, AppStores) it `MUST` state
+its life cycle status according to following rules by any means available
+in the given software distribution solution.
+
+Each release `MUST` clearly state the life cycle
+status and described as: Supported, Deprecated, End of Support.
```
pellared, line 91: If we use "Obsolete" instead of "Deprecated" then we would have distinct status names for "Maturity" and "Releases". What do you think?

specification/versioning.md
```diff
@@ -77,3 +77,35 @@ until the end of that component’s existence:
...
+
+- Latest release `SHOULD` be marked as **Supported** indicating that this is
+  version recommended to new and existing users
+- If applicable latest release of previous `MAJOR` line that is still under
+  active development `MUST` be marked as **Supported** indicating that previous
+  `MAJOR` line is under active development and users not ready to adopt
+  latest `MAJOR` version may use previous versions.
+- Release which was superseded by newer version, thus has entered deprecation
+  period `MUST` be marked as **Deprecated** and `MUST` indicated planned
+  end of support date. This indicates that users should plan for upgrade to
```
pellared, line 106: Do we have already a requirement how long it must be supported? If so can we define it? Otherwise it will be hard to adopt the specification.

What if the latest release fixes a security vulnerability? Shouldn't we immediately drop supportand require a version bump?

specification/versioning.md
```diff
@@ -77,3 +77,35 @@ until the end of that component’s existence:
   MUST be provided as the latest `PATCH` version for the latest `MINOR` version
   of the latest `MAJOR` and SHOULD NOT be provided for previous `PATCH` or
   `MINOR` releases.
+
+## Release life cycle management
+
+Each repository `MUST` use [GitHub releases mechanism](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases)
+to create a versioned snapshot of components and features
```
breedx-splk, line 84:
```suggestion
to publish versioned artifacts of components and features
```

I'd like to avoid "snapshot" here because it has a very specific meaning, at least in Java.

pellared: I updated my suggestion https://github.com/signalfx/gdi-specification/pull/225#discussion_r1159428244 to reflect your proposal

specification/versioning.md
theletterf: I think "Obsolete" is too negative and implies that the thing no longer works, whereas something Deprecated can still work while not being the favored choice.

breedx-splk:
Closing stale PR. If this is still relevant/important, please reopen.

Issue #198, opened 2022-11-14
Handling distribution-specific configuration


pellared:
Currently the spec says:

https://github.com/signalfx/gdi-specification/blob/c9216b2d05faa82da4f5a62d801476994f56ff56/specification/configuration.md?plain=1#L15-L19

In practice, we see that this process it is not working. [Some of the configs in Java OTel distro](https://docs.splunk.com/Observability/gdi/get-data-in/application/java/configuration/advanced-java-otel-configuration.html#metrics-collection-settings) are not defined in the GDI spec, not there are any GitHub issues to introduce them

What should we do?

1. Should we update the GDI configuration spec and introduce the env vars. Like e.g. here: https://github.com/signalfx/gdi-specification/pull/196
2. Should we change change the process to handle distribution-specific configuration?
3. (other proposals)

pellared:
Personally, I propose to allow distributions to have their own config options in the form of `SPLUNK_{LANGUAGE}_{FEATURE}`.

This proposal is based on: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#language-specific-environment-variables

The current spec make sense for configuration that are not language specific. However, we see in practice that some of the configuration is tightly-coupled to some implementation-specific details.

mateuszrzeszutek:
> In practice, we see that this process it is not working.

I think that the current process is unreasonably strict (and this is the reason why it's not being followed everywhere).

Take the profiler, for example: we first developed a POC implementation in Java, and only after we made sure it worked (and that other languages will need this feature too) we made the changes to the GDI spec. The profiler spec was changed several times when other language implementations (like .NET) were being developed.
The current process would require adding everything upfront, before introducing any change to any of the implementation libraries - this simply cannot work.

I think that the GDI spec and the process of adding new things to it should reflect the reality: any behavioral/configuration changes must be added to it IF we want to make them stable, or have more than one instrumentation library implement them.
To put it another way: if it's not in the GDI spec (or OTel spec), a feature cannot be considered stable and might change in backwards incompatible way at any point in time.

pellared:
> The current process would require adding everything upfront, before introducing any change to any of the implementation libraries - this simply cannot work.

[...]

PR #234, opened 2023-06-02
Add a SPLUNK_{COMPONENT}_ prefix recommendation

pellared:
## Why

Fixes https://github.com/signalfx/gdi-specification/issues/198

I want to close the issue and I think that this recommendation should be good enough.

## What

Added:

>  Configuration unique to a specific GDI component (e.g. connected with a language
feature) SHOULD be prefixed with `SPLUNK_{COMPONENT}_` (e.g. `SPLUNK_JAVA_`).

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -31,7 +31,7 @@ exhibited when considering repository-specific configuration variables.
 
 Splunk-specific configuration variables defined in the GDI specification MUST
 be prefixed with `SPLUNK_`. Furthermore, configuration specific to Splunk
-Observability Cloud MUST be prefixed with `SPLUNK_OBSERVABILITY_`, to Splunk
+Observability Cloud MUST be prefixed with `SPLUNK_OBSERVABILITY_` and to Splunk
 Enterprise or Splunk Cloud MUST be prefixed with `SPLUNK_PLATFORM_`.
 Confiugration unique to a specific GDI component (e.g. connected with a lanaguge
 feature) SHOULD be prefixed with `SPLUNK_{COMPONENT}_` (e.g. `SPLUNK_JAVA_`).
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -31,9 +31,11 @@ exhibited when considering repository-specific configuration variables.
 
 Splunk-specific configuration variables defined in the GDI specification MUST
 be prefixed with `SPLUNK_`. Furthermore, configuration specific to Splunk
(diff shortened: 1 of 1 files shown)


Kielek (approved):

specification/configuration.md
```diff
@@ -32,8 +32,10 @@ exhibited when considering repository-specific configuration variables.
 Splunk-specific configuration variables defined in the GDI specification MUST
 be prefixed with `SPLUNK_`. Furthermore, configuration specific to Splunk
 Observability Cloud MUST be prefixed with `SPLUNK_OBSERVABILITY_` and to Splunk
-Enterprise or Splunk Cloud MUST be prefixed with `SPLUNK_PLATFORM_`. If a
-Splunk-specific configuration variable is declared as stable in the GDI
+Enterprise or Splunk Cloud MUST be prefixed with `SPLUNK_PLATFORM_`.
+Configuration unique to a specific GDI component (e.g. connected with a language
+feature) SHOULD be prefixed with `SPLUNK_{COMPONENT}_` (e.g. `SPLUNK_JAVA_`).
```
MrAlias, line 37: Component would be something like a distro, collector, or operator. It seems like the intent here is to scope by language, right?

```suggestion
Configuration unique to a specific programming language (e.g. connected with a language
feature) SHOULD be prefixed with `SPLUNK_{LANGUAGE}_` (e.g. `SPLUNK_JAVA_`).
```

pellared: > Component would be something like a distro, collector, or operator. 

This is correct.

I was initially thinking exactly like you say - only about distributions. However, afterwards I though they can be even Collector specfic env vars `SPLUNK_COLLECTOR_` , so I tried to make a more broad proposal. Also some things may be not connected to language per-se, but related to the ecosystem things that the distribution supports e.g. Micrometer metrics for Java, ASP.NET instrumentation for .NET.

MrAlias: I'm not following, your example is `SPLUNK_JAVA_` not `SPLUNK_COLLECTOR_`. I think this addition needs further clarification.

pellared: `JAVA` stands for Java Distro. I think that your recommendation is more clear and aligned with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#language-specific-environment-variables

pellared:
Closing as my proposal is unclear and I do no think it is worth to spend more our time on it right now.

Issue #242, opened 2023-07-17
Stabilize Real User Monitoring Spans and Attributes


pellared:
https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md#real-user-monitoring-spans-and-attributes needs to be stable for https://github.com/signalfx/gdi-specification/issues/240

pellared:
@t2t2 Can you please take this issue?

PR #245, opened 2023-07-18
Update semantic conventions for RUM

pellared:
Related to https://github.com/signalfx/gdi-specification/issues/242

AFAIK RUM uses OTel Resources which are then send as tags via Zipkin exporter.

From https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md

> This is the only resource attribute that RUM libraries are supposed to set because it's the only one the Zipkin exporter can understand.

I am not sure it this is true.

Also shouldn't we set `splunk.distro.version` instead of `splunk.rum.version`?

@t2t2 @seemk PTAL

PS. Feel free to edit or create a better PR 😉


breedx-splk:
I think we've ripped all the RUM related info out of gdi-spec, so I'm closing this. If I'm mistaken or there's a good reason to keep it, please reopen.

Issue #246, opened 2023-07-18
Real User Monitoring configuration naming


pellared:
From  https://github.com/signalfx/gdi-specification/blob/v1.5.0/specification/configuration.md

> Splunk-specific configuration variables defined in the GDI specification MUST be prefixed with SPLUNK_. 

Therefore shouldn't the all configuration like have a `splunk` prefix e.g.
- `rumAccessToken` -> `splunkAccessToken`
- `realm` -> `splunkRealm`

I know that the section is marked as "Stable". But I believe it is still experimental. See https://github.com/signalfx/gdi-specification/pull/247

breedx-splk:
Actually, after another read though I think I disagree with this assertion. This quote:
> Splunk-specific configuration variables defined in the GDI specification MUST be prefixed with SPLUNK_.

...I think was intending to specifically be referring to environment variables. The two RUM items you call out above are NOT injected through the environment, and I think therefore are not subject to the requirement. 

I still think we should keep it stable.

pellared:
Thanks for your feedback. Closing.

pellared:
@t2t2 @seemk What is your opinion on this? Would you prefer to keep the existing configuration naming? Changing it would probably be annoying for the users. On the other hand, the `splunk` prefix can help in differentiating Splunk specific configuration if OTel defines some standards for RUM (Client Instrumentation).

PS. I am fine closing this issue. It is just a double-check.

seemk:
I guess `splunk*` makes more sense, but can just keep the existing one for backwards compatibility.

pellared:
@seemk Do you want to deprecate existing ones (and still support them) and define new ones?

t2t2:
I don't think it makes as much sense, in case of environment variables `SPLUNK_` prefix ensures there's no conflicts with different vendors expecting their own access tokens, while in RUM case all of the configuration parameters are explicitly passed into splunk libraries - following from documentation examples:

```
// Browser
   SplunkRum.init({
      realm: '<realm>',
      rumAccessToken: '<your_rum_token>',

// IOS
SplunkRum.initialize(beaconUrl: "https://rum-ingest.<realm>.signalfx.com/v1/rum",
      rumAuth: "<rum-token>",

// Android
      SplunkRum.builder()
               .setApplicationName("<name_of_app>")
               .setDeploymentEnvironment("<name_of_env>") // Environment
               .setRealm(realm)
               .setRumAccessToken(rumAccessToken)
```

[...]

PR #180, opened 2022-08-11
Mark RUM configuration items as stable.

breedx-splk:
The configs have not change in 9 months and we are not aware of any pressing/short-term additions to the core configuration.

PR #247, opened 2023-07-18
Mark RUM config as Experimental

pellared:
I know that the section is marked as "Stable". But I believe it is still Experimental.

Related to https://github.com/signalfx/gdi-specification/issues/246

@t2t2 @seemk do you agree?

This would revert https://github.com/signalfx/gdi-specification/pull/180


pellared:
Closing.

Issue #254, opened 2023-09-06
Release v1.6.0


pellared:
Needed for https://github.com/signalfx/gdi-specification/issues/248 as some refinements were made.

Kielek:
https://github.com/signalfx/gdi-specification/releases/tag/v1.6.0

PR #256, opened 2023-09-12
Fix permissions to be applicable for non-instrumentation repositories

pellared:
Blocker for https://github.com/signalfx/gdi-specification/issues/254

The current repository permissions does not make sense for repos such us https://github.com/signalfx/splunk-otel-collector, https://github.com/signalfx/splunk-otel-collector-operator, https://github.com/signalfx/splunk-otel-collector-chart, https://github.com/signalfx/splunk-otel-collector-heroku

### DIFF
--- a/specification/repository.md
+++ b/specification/repository.md
@@ -11,6 +11,10 @@ approval is granted, GDI repositories MUST NOT cut a GA release.
 
 ### Teams
 
+- MUST have an admins team and teams MAY be shared between repositories
+  - MUST have `-admins` suffix
+  - MUST include at least two currently full-time Splunkers
+  - MUST NOT include any non-full-time Splunkers
 - MUST have a maintainers team and teams MAY be shared between repositories
   - MUST have `-maintainers` suffix
   - MUST include at least two currently full-time Splunkers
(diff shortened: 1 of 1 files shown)


pellared:
@mateuszrzeszutek @Kielek PTAL

breedx-splk:
Can you elaborate on why you think we need an additional "admin" team? Aren't the maintainers basically admins?

pellared:
> Can you elaborate on why you think we need an additional "admin" team? Aren't the maintainers basically admins?

We have `gdi-instrumentation-admins` team which is added to most (all?) our GDI instrumentation repositories.

Kielek (approved):

mateuszrzeszutek (approved):
I think I'm fine with that, although I'd prefer the doc to say "at least one of these two teams" explicitly.

pellared:
@mateuszrzeszutek Like:

```
- MUST grant `Admin` [permission
  level](https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-permission-levels-for-an-organization)
  to the admins and/or maintainers team
```

?

> I think I'm fine with that, although I'd prefer the doc to say "at least one of these two teams" explicitly.

I think we can refine this in a bugfix release. You can also propose a change if you already have an idea how to define it.

PR #263, opened 2023-09-27
Status and Support Level

Kielek:
Definitions prepared by @akubik-splunk.


specification/status_and_support_level.md
```diff
@@ -0,0 +1,30 @@
...
+element or artifact of the component, or feature. Status definitions help set
+expectations around upcoming changes and related levels of support.
+
+| Level        | Development stage | Description                                                                                                                                                                                                                                                                                               |
+|--------------|-------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Stable       |                   | The implementation of rhe distribution, component, or element is complete in a defined scope and should not change as part of the current major version line unless a critical bug or security issue requires it.                                                                                         |
```
theletterf, line 13:
```suggestion
| Stable       |                   | The implementation of the distribution, component, or element is complete in a defined scope and should not change as part of the current major version line unless a critical bug or security issue requires it.                                                                                         |
```

theletterf (approved):

akubik-splunk:
For our own protection this file needs to contain headers included in  the source document. Especially it has to directly refer to Splunk Terms and Conditions link as main source of definitions.

specification/status_and_support_level.md
```diff
@@ -0,0 +1,30 @@
+# Status and support levels of open source components
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+## Definitions of status levels
```
pellared, line 5: The status levels definitions are scattered in two places (here and `specification/versioning.md`). It makes me hard to follow.
I think that there are 3 concepts and they should be described in the following order:
  - status levels and development stages of components
  - support levels depending on the component owner and status level
  - versioning (and structuring) depending on component status level
 
I propose to describe everything above in a file named  `component-status.md` or simply `status.md`. I think we can even get rid of `versioning.md` as I do not think we need to call out the specification versioning policy.

pellared: Based on the conversion it should be spitted into 3 documents:
- `status.md`
- `support.md`
- `versioning.md`

In future we will also need `releasing.md` defining the component releasing requirements.
Changing the PR to draft.

Kielek:
@akubik-splunk, I am closing this PR as a stale. Feel free to reopen when needed.

PR #284, opened 2023-12-06
Clarify default sampler

seemk:
OpenTelemetry defaults to `parentbased_always_on`, but this is not consistent across our distributions:
* `parentbased_always_on` is used by Node.js, .NET, Python
* `always_on` is used by Java, [Go](https://github.com/signalfx/splunk-otel-go/blob/4b4f13d128bffd4035331cb84dc796a07b448803/distro/otel.go#L159)

If we are advertising as a no sampling solution, we should default to `always_on`, but still allow for customers to change it.
Using `parentbased_always_on` can cause data loss if customers have previously set up cloud tracing, which injects a `traceparent` with `00` for `traceflags`.

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -242,6 +242,8 @@ are required.
     OTLP exporter instead, or set the SPLUNK_REALM and SPLUNK_ACCESS_TOKEN environment variables to send 
     telemetry directly to Splunk Observability Cloud.
     ```
+- `OTEL_TRACES_SAMPLER`
+  - Distribution MUST default to `always_on`.
 
 In addition to environment variables, other ways of defining configuration also exist:
 
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -242,6 +242,7 @@ are required.
     OTLP exporter instead, or set the SPLUNK_REALM and SPLUNK_ACCESS_TOKEN environment variables to send 
     telemetry directly to Splunk Observability Cloud.
     ```
+
 - `OTEL_TRACES_SAMPLER`
   - Distribution MUST default to `always_on`.
 
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -244,7 +244,8 @@ are required.
     ```
 
 - `OTEL_TRACES_SAMPLER`
-  - Distribution MUST default to `always_on`.
+  - Distribution MUST default to `always_on`
+    (not OpenTelemetry default)
 
 In addition to environment variables, other ways of defining configuration also exist:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,13 @@
 
 #### Enhancements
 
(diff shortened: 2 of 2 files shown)


seemk:
@akubik-splunk

specification/configuration.md
```diff
@@ -243,6 +243,9 @@ are required.
     telemetry directly to Splunk Observability Cloud.
     ```
 
+- `OTEL_TRACES_SAMPLER`
+  - Distribution MUST default to `always_on`.
```
Kielek, line 247:
```suggestion
  - Distribution MUST default to `always_on`
     (not OpenTelemetry default)
```

pellared:
> In which distros this is marked as stable?

For sure in [.NET](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#samplers) it is stable.

In Python it is part of its [SDK](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python%20OTEL_TRACES_SAMPLER&type=code) which is stable.

akubik-splunk:
> > In which distros this is marked as stable?
> 
> For sure in [.NET](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#samplers) it is stable.
> 
> In Python it is part of its [SDK](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python%20OTEL_TRACES_SAMPLER&type=code) which is stable.

So it seems to me that we should prep for the major bump and if possible bundle it up with other changes e.g. HTTP semantics :) for .NET and for Python as well. If there are other major changes that we would like to do let's use this as opportunity to do so. We would target such changes for the Jan-Feb next year though

pellared (approved):

pellared:
> It is against OTel spec defaults.

Correct, this is why we want to change the default in the GDI spec.

>  We need agreement if we really should update it in all Splunk distros.

I think we should. We intentionally decided to use `always_on` in Go to address the issue from the PR description:

> If we are advertising as a no sampling solution, we should default to `always_on`, but still allow for customers to change it.
Using `parentbased_always_on` can cause data loss if customers have previously set up cloud tracing, which injects a `traceparent` with `00` for `traceflags`.

AFAIK Node.js plans to do the change in their upcoming v3 release. cc @seemk 
I think it would be good to do the same in Python and .NET distributions.
Personally, I would not require a major bump for such change .

@akubik-splunk, do you have an opinion here?

Kielek (approved):
I do not like the idea changing default behavior from OTel. It can be problematic, especially in terms of promoting upstream distributions (for non-most-popular technologies.

Approving, as you agreed on this.

pellared:
> especially in terms of promoting upstream distributions

One can always set the `always_on` while using upstream. This is just for make it more convenient for our users.

PR #335, opened 2025-02-11
Add integration context spec

breedx-splk:
In order to facilitate better interoperability between AppDynamics and Splunk O11y Cloud, we need to pass some additional context. This is the first step in specifying which fields the splunk-otel side of things will need to consume and to generate. 

The AppD side is intended to receive a follow-up PR (although one could make the case that the AppD side shouldn't be spec'd here....I'm open to ideas).


specification/tracestate.md
```diff
@@ -0,0 +1,75 @@
...
+| `appd_acct_id`   | string | `appd.upstream.account_id` | The AppDynamics account ID of the upstream component making the request | 65230, 10018b                     |
+| `appd_app_id`    | string | `appd.upstream.app_id`     | The ID of the AppDynamics instrumented application.                     | 0293845, destination-factory-9000 |
+| `appd_tier_id`   | string | `appd.upstream.tier_id`    | The "tier id" to which the AppDynamics instrumented application belongs | 12, xdev.tier9                    |
+
+All other `tracestate` fields SHOULD be ignored and not placed into span
+attributes, unless required elsewhere.
+
+## Outgoing State
+
+When `cisco.tracestate.enabled` is `true`, Splunk implementations MUST
```
Kielek, line 61: AppD instrumented App --- passing `tracestate` ---> Splunk/OTel instrumented App ----- Should we also pass received `tracestate` from the parent AppD app?-----> AppD instrumented App

specification/behaviors.md
```diff
@@ -46,3 +46,8 @@ Logs](https://github.com/open-telemetry/opentelemetry-specification/tree/main/sp
 The logs containing profiling data MUST be sent via OTLP. Instrumentation
 libraries SHOULD reuse persistent OTLP connections from other signals (traces,
 metrics).
+
+## Trace State Interchange
```
johnbley, line 50: We discussed this in person, but have you given any more thought to the idea of using one or more separate http headers rather than `tracestate`?  This would seem to fit the "one hop" requirement much better than abusing the "trace-wide" `tracestate`.  If not, then you need to spell out the "receive and delete" one-hop nature of these fields (without which the ability to link between the product sets gets horribly confused)

johnbley (changes requested):

specification/tracestate.md
```diff
@@ -0,0 +1,75 @@
...
+The remainder of this document applies only when the W3C Trace Context
+propagator is active. Setting `cisco.tracestate.enabled` to `true` SHOULD
+automatically enable the W3C Trace Context propagator, and MAY override
+other existing configuration/settings. If the W3C Trace Context propagator
+cannot be enabled, then the rest of this document is irrelevant and can
+be ignored.
+
+## Incoming State
+
+When `cisco.tracestate.enabled` is `true`, Splunk implementations MUST
```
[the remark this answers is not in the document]
breedx-splk: Probably not. Sorry.

specification/behaviors.md
breedx-splk: Uh. I was attempting (probably poorly) to remain consistent with other design docs I started with. Those docs used the word `traceState` which I took to mean the actual `tracestate` spec/header, but I guess I'm mistaken....or more likely that's just a poor choice when some parties aren't doing tracing anyway.

Let's just cook up something else custom then.

specification/tracestate.md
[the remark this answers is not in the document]
breedx-splk: Going to ditch tracestate in favor of a custom thing so it won't be relevant, but thanks for pointing it out!

breedx-splk:
Going to continue working on this now, without using `tracestate` or w3c, but instead some custom headers...

specification/behaviors.md
breedx-splk: migrating to separate headers now.

specification/tracestate.md
breedx-splk: Well we've drifted away from `tracestate` now for these, but your question stands....do we need some of the incoming state _propagated_ to the outgoing state? I think the answer to this likely depends on how sophisticated the backend integrations will be and how deep the links may go. We should find out.

specification/integration_context.md
```diff
@@ -0,0 +1,86 @@
...
+attributes to any Spans created as part of the incoming request context.
+Null or missing values MUST be handled gracefully by simply
+omitting the span attributes.
+
+| header              | span attribute             | description                                                             | example                           |
+|---------------------|----------------------------|-------------------------------------------------------------------------|-----------------------------------|
+| `cisco-ctx-acct-id` | `appd.upstream.account_id` | The AppDynamics account ID of the upstream component making the request | 65230, 10018b                     |
+| `cisco-ctx-app-id`  | `appd.upstream.app_id`     | The ID of the AppDynamics instrumented application.                     | 0293845, destination-factory-9000 |
+| `cisco-ctx-bt-id`   | `appd.upstream.bt_id`      | The ID of the upstream AppDynamics business transaction (BT)            | 209834098273                      |
+| `cisco-ctx-tier-id` | `appd.upstream.tier_id`    | The "tier id" to which the AppDynamics instrumented application belongs | 12, xdev.tier9                    |
```
Kielek, line 68: OTel semantic convention tends to avoid `smth_id`. It is better to keep `smth` as a namespace in case oneeding to add `smth.super-important-new-tag`

```suggestion
| `cisco-ctx-acct-id` | `appd.upstream.account.id` | The AppDynamics account ID of the upstream component making the request | 65230, 10018b                     |
| `cisco-ctx-app-id`  | `appd.upstream.app.id`     | The ID of the AppDynamics instrumented application.                     | 0293845, destination-factory-9000 |
| `cisco-ctx-bt-id`   | `appd.upstream.bt.id`      | The ID of the upstream AppDynamics business transaction (BT)            | 209834098273                      |
| `cisco-ctx-tier-id` | `appd.upstream.tier.id`    | The "tier id" to which the AppDynamics instrumented application belongs | 12, xdev.tier9                    |
```

specification/integration_context.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Incoming State
+
+When `cisco.ctx.enabled` is `true`, Splunk implementations MUST
+extract fields from the `cisco-ctx-*` headers (above) and add extra
+attributes to any Spans created as part of the incoming request context.
+Null or missing values MUST be handled gracefully by simply
+omitting the span attributes.
+
+| header              | span attribute             | description                                                             | example                           |
```
Kielek, line 63: I think that, you are missing here span type. Part of them seems to be pure numbers, should we treat them as int/strings?

specification/integration_context.md
```diff
@@ -0,0 +1,86 @@
...
+configuration SHOULD be named `CISCO_CTX_ENABLED`. The default value
+MUST be `false` or equivalent language-specific non-truthy value.
+
+## Incoming State
+
+When `cisco.ctx.enabled` is `true`, Splunk implementations MUST
+extract fields from the `cisco-ctx-*` headers (above) and add extra
+attributes to any Spans created as part of the incoming request context.
+Null or missing values MUST be handled gracefully by simply
+omitting the span attributes.
```
Kielek, line 61: Before merging, we should create POC for this.
I do no like the idea of adding new attributes to every span. Why not just to the first span in the application? Other calculation could be done by the standard trace-id, span-id and the hierarchy.

Kielek (commented):
Did you considered [W3C baggage](https://www.w3.org/TR/baggage/)?

I didn't have time to full analysis, but it is something we should consider.

specification/integration_context.md
```diff
@@ -0,0 +1,86 @@
...
+
+| header              | span attribute             | description                                                             | example                           |
+|---------------------|----------------------------|-------------------------------------------------------------------------|-----------------------------------|
+| `cisco-ctx-acct-id` | `appd.upstream.account_id` | The AppDynamics account ID of the upstream component making the request | 65230, 10018b                     |
```
johnbley, line 65: Might be worth calling out that `.upstream.` is chosen specifically to avoid collisions with the span attributes added  by the dual-signal agent (when enabled)?

specification/integration_context.md
johnbley: Yeah, worth calling out that we should treat them as opaque strings.

specification/integration_context.md
```diff
@@ -0,0 +1,86 @@
...
+# Integration Context
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+In order to create a more complete picture of a distributed environment, we
+propose the use of additional headers to pass extra context details between
+Cisco observability components. Specifically, we are concerned with passing
+information between OpenTelemetry-based Splunk distributions and AppDynamics
+agents, in both directions.
+
```
johnbley, line 18: Providing product linkage in both directions, not passing data in both directions on the http request/response.  (Headers below are all request headers).

breedx-splk: Are you suggesting that I add some additional details about the purpose of passing data then here? I'm not sure I understand your comment.

breedx-splk:
> Did you considered [W3C baggage](https://www.w3.org/TR/baggage/)?

Yes. And am confident that it would work, especially since [mutations are allowed](https://www.w3.org/TR/baggage/#mutating-baggage). I'm not sure how easy/sophisticated the baggage handling capabilities are in the various languages though.

I would be happy to convert these headers to baggage keys. What do other folks think?

specification/integration_context.md
breedx-splk: I think it needs to be on every span for that request. It is more concise that way. 

You also have to allow for a scenario where a call enters the app, leaves the app, and comes back in from a different tier. Multiple tiers possibly for the same trace from different tiers.

You said that you don't like adding attributes, but you didn't say why.

MrAlias:
:whale:

specification/integration_context.md
breedx-splk: sure, agreed. added it above.

specification/integration_context.md
breedx-splk: Sure, that helps to clarify. Added.

johnbley:
> > Did you considered [W3C baggage](https://www.w3.org/TR/baggage/)?
> 
> Yes. And am confident that it would work, especially since [mutations are allowed](https://www.w3.org/TR/baggage/#mutating-baggage). I'm not sure how easy/sophisticated the baggage handling capabilities are in the various languages though.
> 
> I would be happy to convert these headers to baggage keys. What do other folks think?

Unless I misunderstand something, I'm against using baggage for this because baggage is meant to be trace-wide and we'd have to write logic to counteract that. Yes, it is possible, but it's much easier (less code, less compexity, fewer chances for errors or misattribution) when it is a plain non-propagated header.

breedx-splk:
> > > Did you considered [W3C baggage](https://www.w3.org/TR/baggage/)?
> baggage is meant to be trace-wide and we'd have to write logic to counteract that. Yes, it is possible, but it's much easier (less code, less compexity, fewer chances for errors or misattribution) when it is a plain non-propagated header.

Right, we'd also need to make sure that we remove some fields at each hop, because the _default_ behavior is to propagate, which is not necessarily what we want/need here. Given that the intent of baggage is to propagate, I think it suggests that our use case is slightly different.

Thanks @johnbley .

specification/tracestate.md
johnbley: Just the BT ID, which could/should be added to baggage in a clean way.  Everything else (as far as I know) is meant to be one-hop and interpreted as such.

johnbley: That would make the most sense to me unless there's something about bt.id I don't understand,.

specification/integration_context.md
johnbley: Just trying to clarify that the phrase "passing information... in both directions" might suggest that information is passed on http request and response, when it will not be.  Not a big deal.

breedx-splk:
Prototype here https://github.com/signalfx/splunk-otel-java/pull/2198

johnbley (approved):

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+* `cisco-ctx-tier-id` - Contains the ID of the AppDynamics tier.
+* `cisco-ctx-bt-id` - Contains the ID of the AppDynamics business transaction (BT).
+* `cisco-ctx-env` - Contains
+  the [deployment.environment.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/)
+  resource value from an OpenTelemetry based component.
+* `cisco-ctx-service` - Contains the [service.name](https://opentelemetry.io/docs/specs/semconv/resource/#service)
+  resource value from an OpenTelemetry based component.
+
+HTTP headers are capable of being multivalued. As such, implementations
+SHOULD use the _last_ value when the above headers contain multiple values.
```
pellared, line 37: I think that the common pattern is actually to use the first value :sweat_smile: 
This is also reflected in (most?) standard libraries. E.g.
- Java `getFirst`: https://docs.oracle.com/javase/8/docs/jre/api/net/httpserver/spec/com/sun/net/httpserver/Headers.html
- Go `Get`: https://pkg.go.dev/net/http#Header.Get
- Rust: https://docs.rs/http/latest/src/http/header/map.rs.html#730-756

Some background:
- https://github.com/golang/go/issues/51493#issuecomment-1059588158

Moreover, this header controlled by us so producing multiple values for the same header is already a bug. Long story short, it should never happen. Therefore, it does not matter if we take the first or last as we have no guarantee which value is "better".

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+* `cisco-ctx-app-id` - Contains the ID of the AppDynamics application.
+* `cisco-ctx-tier-id` - Contains the ID of the AppDynamics tier.
+* `cisco-ctx-bt-id` - Contains the ID of the AppDynamics business transaction (BT).
+* `cisco-ctx-env` - Contains
+  the [deployment.environment.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/)
+  resource value from an OpenTelemetry based component.
+* `cisco-ctx-service` - Contains the [service.name](https://opentelemetry.io/docs/specs/semconv/resource/#service)
+  resource value from an OpenTelemetry based component.
+
+HTTP headers are capable of being multivalued. As such, implementations
```
pellared, line 36: I think we should refer to https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2:

> A sender MUST NOT generate multiple header fields with the same field name in a message unless either the entire field value for that header field is defined as a comma-separated list [i.e., #(values)] or the header field is a well-known exception (as noted below).

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+agents, in both directions.
+
+# Headers
+
+The following defines several new headers and describes their intended use.
+All headers MUST be treated as optional -- peer services will not always
+generate them. These headers SHOULD be treated as opaque values of type
+string.
+
+* `cisco-ctx-acct-id` - Contains the ID of the AppDynamics account.
```
pellared, line 26: Maybe `AppD` prefix as it feels more unique than `cisco`?

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+* `cisco-ctx-app-id` - Contains the ID of the AppDynamics application.
+* `cisco-ctx-tier-id` - Contains the ID of the AppDynamics tier.
+* `cisco-ctx-bt-id` - Contains the ID of the AppDynamics business transaction (BT).
+* `cisco-ctx-env` - Contains
+  the [deployment.environment.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/)
+  resource value from an OpenTelemetry based component.
+* `cisco-ctx-service` - Contains the [service.name](https://opentelemetry.io/docs/specs/semconv/resource/#service)
+  resource value from an OpenTelemetry based component.
```
pellared, line 34: nit: can we use the "standardized/canonical" way of defining headers

```suggestion
* `Cisco-Ctx-Acc-ID` - Contains the ID of the AppDynamics account.
* `Cisco-Ctx-App-ID` - Contains the ID of the AppDynamics application.
* `Cisco-Ctx-Tier-ID` - Contains the ID of the AppDynamics tier.
* `Cisco-Ctx-BT-ID` - Contains the ID of the AppDynamics business transaction (BT).
* `Cisco-Ctx-Env` - Contains
  the [deployment.environment.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/)
  resource value from an OpenTelemetry based component.
* `Cisco-Ctx-Service` - Contains the [service.name](https://opentelemetry.io/docs/specs/semconv/resource/#service)
  resource value from an OpenTelemetry based component.
```

pellared (commented):
Some comments. I have not managed to review everything, but looks good in general.

Kielek (commented):
@johnbley , @breedx-splk, whats about https://github.com/signalfx/gdi-specification/pull/335#discussion_r1961025413?

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+Cisco observability components. Specifically, we are concerned with passing
+information between OpenTelemetry-based Splunk distributions and AppDynamics
+agents, in both directions.
+
+# Headers
+
+The following defines several new headers and describes their intended use.
+All headers MUST be treated as optional -- peer services will not always
+generate them. These headers SHOULD be treated as opaque values of type
+string.
```
Kielek, line 24: What means opaque here?
Should internal.public API implementing it treat it as an additional class with string value field, liek the Baggage + metadata?

pellared: This sentence should be indeed clarified. There is too much left for interpretation.

I think the intention is to have a dedicated opaque type for each header.
Thanks to it in future you can add funtionality to use  `cisco-ctx-app-id` as `int` and leave other intact.

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#set-value

>  `Metadata` [..] This should be an opaque wrapper for a string with no semantic meaning. Left opaque to allow for future functionality.

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+* `cisco-ctx-env` - Contains
+  the [deployment.environment.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/)
+  resource value from an OpenTelemetry based component.
+* `cisco-ctx-service` - Contains the [service.name](https://opentelemetry.io/docs/specs/semconv/resource/#service)
+  resource value from an OpenTelemetry based component.
+
+HTTP headers are capable of being multivalued. As such, implementations
+SHOULD use the _last_ value when the above headers contain multiple values.
+
+## Splunk OpenTelemetry distributions
```
pellared, line 39: I think it would be helpful to describe how this can be implemented using OTel components (propagators, processors) based on https://github.com/signalfx/splunk-otel-java/pull/2198

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+consumption and propagation of additional Cisco (bespoke) integration context.
+
+The configuration SHOULD be named `cisco.ctx.enabled` or an idiomatic
+equivalent for each language. If using environment variables, the
+configuration SHOULD be named `CISCO_CTX_ENABLED`. The default value
+MUST be `false` or equivalent language-specific non-truthy value.
+
+## Incoming State
+
+When `cisco.ctx.enabled` is `true`, Splunk implementations MUST
```
pellared, line 58:
```suggestion
When `cisco.ctx.enabled` configuration is `true`, Splunk implementations MUST
```

Just for clarity to not make a mistake that it is a header :sweat_smile:

specification/integration_context.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Configuration
+
+Splunk OTel distributions MUST provide a way for users to opt-into the
+consumption and propagation of additional Cisco (bespoke) integration context.
+
+The configuration SHOULD be named `cisco.ctx.enabled` or an idiomatic
+equivalent for each language. If using environment variables, the
+configuration SHOULD be named `CISCO_CTX_ENABLED`. The default value
+MUST be `false` or equivalent language-specific non-truthy value.
```
pellared, line 54: Are we sure that it should be disabled by default? Would it not be inconvenient for users to enable it?
Do we even need such configuration?
If a AppD integration/communication would not be present than it would be an "no-op". Requiring to enable may be inconvenient.

Kielek: I think that it might be needed for outgoing requests.

pellared: I guess the integration somehow still needs to find out which "outgoing requests" are targeting AppD. Otherwise, the headers would leak in all outgoing requests.

specification/integration_context.md
laurit: to me opaque here means that agents should not attempt to parse or otherwise derive a meaning for these headers, they are just values as far as agent is concerned

specification/integration_context.md
laurit: Using the first value is much easier. Note that implementation in https://github.com/signalfx/splunk-otel-java/pull/2198 also uses first value.

specification/integration_context.md
laurit: > Do I understand correctly that the propagator should be applied only in communication between Splunk and AppD?

How would one tell that context is being propagated to an app that is using appd? There is an open spec issue for controlling context propagation boundary. I hope that none of these headers are considered sensitive as limiting where they are propagated will be complicated.

johnbley: @laurit raises a good point that a customer might want the control of this behavior to be (optionally) more granular - e.g. `cisco.ctx.enabled`would be the preferred way, but also `cisco.ctx.enabled.inbound=true/false` and `cisco.ctx.enabled.outbound` as more specific toggles.  (If so, further spec issue to settle - what if the "combination" `enabled` is set and a more specific one is also set - which takes precedence?)

johnbley: For outbound injection, instrumentation is obviously not going to be able to tell what it's talking to.  For security/leakage concerns I've suggested more granular behavior controls, but the fallback could also be telling the user to configure a network proxy/firewall/etc. for their needs.  

This feature (appd<->splunk context sharing) needs to be off by default in `splunk-otel` per PM, who I'm sure would be  happy to share reasoning in separate channels.

specification/integration_context.md
breedx-splk: That's cool, appreciate that. I'll change it. 👍🏻

specification/tracestate.md
breedx-splk: Putting some parts over here and some parts over there makes it slightly more difficult to wrangle. So far, it feels fairly simple/straightforward to me. I don't feel super strongly about it.

specification/integration_context.md
breedx-splk: Thanks for clarifying. I tweaked it and remove the mention of "both directions". 👍🏻

specification/integration_context.md
breedx-splk: What I like about `cisco` is that we are able to use the same prefix for data generated in AppD components and data generated in splunk-otel components. With this approach, the intention is to be able to say "Oh, I see this `cisco-ctx-*` header, I know exactly why it exists -- to facilitate integrated product experience.

specification/integration_context.md
breedx-splk: I don't think it matter much -- but my purely aesthetic preference is to use lowercase. Most libraries/frameworks standardize on lowercase anyway I believe.

specification/integration_context.md
breedx-splk: > to me opaque here means that agents should not attempt to parse or otherwise derive a meaning for these headers, they are just values as far as agent is concerned

@laurit is exactly right -- the intent is that it's of type string and that it shouldn't need to be parsed. I think the use of "opaque" to mean this is common.

breedx-splk: > I think the intention is to have a dedicated opaque type for each header.
Thanks to it in future you can add funtionality to use cisco-ctx-app-id as int and leave other headers intact.

No, I don't want any of this spec'd as anything other than a string.

specification/integration_context.md
breedx-splk: > Do I understand correctly that the propagator should be applied only in communication between Splunk and AppD?

No, this was not the intention. What phrasing leads you to believe this?

> How would one tell that context is being propagated to an app that is using appd?

You can't and shouldn't need to. The headers shouldn't contain sensitive information. Users who choose to put sensitive information in any of these fields (service name, environment, appd IDs, whatever) shouldn't use this feature.

specification/integration_context.md
breedx-splk: It's fine that, when enabled, the headers are in outgoing requests. That's intentional. 

What the default of `false` does is prevent non-AppD users from getting these unnecessary headers. Even some AppD users probably don't want it, so the default state should maintain the status quo. It takes extra config/effort to turn on the special/new feature. Yes.

specification/integration_context.md
pellared: > Most libraries/frameworks standardize on lowercase anyway I believe.

I am not sure. For sure this is not true for Go :laughing: 
- https://pkg.go.dev/net/http#Header
- https://pkg.go.dev/net/http#CanonicalHeaderKey

> I don't think it matter much 

It does not, but this is how most RFC and docs like https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers follow :wink: 

I won't fight for it, but sometime the difference in "casing" makes it easier for me to "pattern-match" the "type". E.g. when I see `SPLUNK_IS_COOL` I see env var :laughing:

specification/integration_context.md
pellared: >  the intent is that it's of type string and that it shouldn't need to be parsed. I think the use of "opaque" to mean this is common.

I strongly disagree. Opaque is more like encapsulation and information hiding.

I just found https://en.wikipedia.org/wiki/Opaque_data_type to back up my understanding.

Also from Oxford Dictionary:

> opaque - (adj.) not able to be seen through; not transparent.

specification/integration_context.md
pellared: > What phrasing leads you to believe this?

This:

> See [integration_context.md](integration_context.md) for specifics about
exchanging additional context between AppD and splunk-otel based agents.

Threat 1:  Information disclosure
If I understand correctly the current design sends splunk-otel context to everyone (via outgoing requests).

Thread 2: Spoofing, Tampering?
The other problem is that the incoming requests are also not verifying who send the data (if this is actually coming from AppD). A malicious actor can send `cisco-ctx-acct-id` headers which would be stored in spans.

specification/integration_context.md
pellared: Resolving per this conversation which I see as a duplicate: https://github.com/signalfx/gdi-specification/pull/335#discussion_r1965156593

specification/integration_context.md
breedx-splk: Ok, rather than getting hung up on this then, I'll just remove this section and we can assume that the header won't be duplicated.

specification/integration_context.md
breedx-splk: Resolving and leaving as lowercase for now. Can revisit if there is increased pressure sometime. I also don't feel super strongly about it.

specification/integration_context.md
breedx-splk: > Then let's spec it as "of string type"

As I wrote it, it literally says "opaque values of type string" but you seem hung up on the use of "opaque". 

Anyway, I like the summary in this stack overflow post:
image

at least that is the exact same sentiment I was aiming for. It's just some characters, don't parse it, don't extract meaning from those bytes. Just pass it along.

specification/integration_context.md
breedx-splk: While yes, technically, this approach of passing data in headers does disclose information, I don't think any of these fields can be considered sensitive. 

And I'm not sure what else to document -- this spec here is just describing what data goes in headers. If you have a service name or environment that is hyper secretive and you don't want it disclosed, sure, you shouldn't use this feature. However, I'm not sure that documenting it _here_ would help any end users of splunk-otel-* or appd agents. 

In hopes of moving this forward, I will add a note.

specification/integration_context.md
pellared: I am fine with the current wording. I know realized that it is in the used in the context where we talk about the meaning of the header values and not how we want to design/define types in our distros.

@Kielek, PTAL

pellared (approved):

PR #336, opened 2025-02-17
Define `profiling.instrumentation.source` Log Message Attribute

tduncan:
The profiling backend ingestion process needs to be able to distinguish between callstacks collected by the existing continuous "Always On" profiler and those collected by the upcoming trace snapshot profiler. This PR defines an attribute to be included in a log message body that will identify which profiling process was responsible for collecting and reporting the exported callstacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Semantic Conventions
+
+#### Enhancements
+
+- Define the `PROFILING_INSTRUMENTATION_SOURCE` attribute with valid values of `continuous` and `snapshot`
+
 ## [1.7.0] - 2025-01-07
 
 ### Configuration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 #### Enhancements
 
-- Define the `PROFILING_INSTRUMENTATION_SOURCE` attribute with valid values of `continuous` and `snapshot`
+- Define the `profiling.instrumentation.source` attribute with valid values of `continuous` and `snapshot`.
+[#336](https://github.com/signalfx/gdi-specification/pull/336)
 
 ## [1.7.0] - 2025-01-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 #### Enhancements
 
-- Define the `profiling.instrumentation.source` attribute with valid values of `continuous` and `snapshot`.
+- Define the `profiling.instrumentation.source` attribute with valid values
+  of `continuous` and `snapshot`.
 [#336](https://github.com/signalfx/gdi-specification/pull/336)
 
 ## [1.7.0] - 2025-01-07
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -129,7 +129,10 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
-- `profiling.instrumentation.source` MUST be set to either `continuous` or `snapshot`
+- `profiling.instrumentation.source` MUST be set to either `continuous` or `snapshot` to 
+  identify the profiling mechanism reporting the callstacks
+  - `continuous` SHOULD be used to indicate the continuous "Always On" profiling
+  - `snapshot` SHOULD be used to indicate the trace snapshot profiling
 
 #### `LogRecord` Message `text` Data Format Specific Attributes
 
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -129,6 +129,7 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
+- `profiling.instrumentation.source` MUST be set to either `continuous` or `snapshot`
 
 #### `LogRecord` Message `text` Data Format Specific Attributes
 
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -129,10 +129,9 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
-- `profiling.instrumentation.source` MUST be set to either `continuous` or `snapshot` to 
-  identify the profiling mechanism reporting the callstacks
-  - `continuous` SHOULD be used to indicate the continuous "Always On" profiling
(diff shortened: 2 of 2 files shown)


pellared (commented):
Can you please add a changelog entry?

specification/semantic_conventions.md
```diff
@@ -129,6 +129,7 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
+- `profiling.instrumentation.source` MUST be set to either `continuous` or `snapshot`
```
breedx-splk, line 132: I think we should add a few more words here that describe _when_ one or the other should be used and briefly what they mean. Like literally one or two sentences.

tduncan:
> Can you please add a changelog entry?

Done.

specification/semantic_conventions.md
tduncan: Additional context added. Feedback is very welcome!

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Semantic Conventions
+
+#### Enhancements
+
+- Define the `PROFILING_INSTRUMENTATION_SOURCE` attribute with valid values of `continuous` and `snapshot`
```
pellared, line 9:
```suggestion
- Define the `profiling.instrumentation.source` attribute with valid values of `continuous` and `snapshot`.
[#336](https://github.com/signalfx/gdi-specification/pull/336)
```

specification/semantic_conventions.md
```diff
@@ -129,6 +129,10 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
+- `profiling.instrumentation.source` MUST be set to either `continuous` or `snapshot` to 
+  identify the profiling mechanism reporting the callstacks
+  - `continuous` SHOULD be used to indicate the continuous "Always On" profiling
+  - `snapshot` SHOULD be used to indicate the trace snapshot profiling
```
pellared, line 135: how about

```suggestion
- `profiling.instrumentation.source` MUST be set to either:
  - `continuous` for continuous "Always On" profiling
  - `snapshot` for trace snapshot profiling
```

breedx-splk (approved):

pellared (approved):

specification/semantic_conventions.md
```diff
@@ -129,6 +129,9 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
+- `profiling.instrumentation.source` MUST be set to either:
+  - `continuous` for continuous "Always On" profiling
+  - `snapshot` for trace snapshot profiling
```
Kielek, line 134: I think that, for backward compatibility, you should mention that it will be treated as `continuous`.

Kielek (commented):
I am not against change for a short term solution, but we need to start thinking how to handle the OTLP profiling protocol. I hope it will go live this year.

In such cases you will be not able to expect such attributes.
For known, libraries, you can probably utilize 
 * [InstrumentationScope](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L138)
For all generic cases:
* [SampleType](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L192)
* [PeriodType](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L225)

specification/semantic_conventions.md
seemk: This, there are too many existing installations, my suggestion is that `profiling.instrumentation.source` must only be set when anything other than `continuous` is used (as the backend treats all of the profiles as continuous by default).

tduncan:
> I am not against change for a short term solution, but we need to start thinking how to handle the OTLP profiling protocol. I hope it will go live this year.
> 
> In such cases you will be not able to expect such attributes. For known, libraries, you can probably utilize
> 
> * [InstrumentationScope](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L138)
>   For all generic cases:
> * [SampleType](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L192)
> * [PeriodType](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L225)

`SampleType` and `PeriodType` would seem to be describing the data itself, where as what we need to know is information about the context from which the callstacks were collected.

I was hoping to use `InstrumentationScope` but conversations with individuals more familiar with the GDI process made it clear that, at least for now, the `InstrumentationScope` information will not be available for us to rely on.

Kielek:
> > I am not against change for a short term solution, but we need to start thinking how to handle the OTLP profiling protocol. I hope it will go live this year.
> > In such cases you will be not able to expect such attributes. For known, libraries, you can probably utilize
> > 
> > * [InstrumentationScope](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L138)
> >   For all generic cases:
> > * [SampleType](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L192)
> > * [PeriodType](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L225)
> 
> `SampleType` and `PeriodType` would seem to be describing the data itself, where as what we need to know is information about the context from which the callstacks were collected as the values for both `SampleType` and `PeriodType` would be identical. The shape of the profiling data does not vary between the existing continuous profiling solution and the trace snapshot extension we're porting into the Java agent.
> 
> I was hoping to use `InstrumentationScope` but conversations with individuals more familiar with the GDI process made it clear that, at least for now, the `InstrumentationScope` information will not be available for us to rely on.

For our internal internal profiling format, we can put any data we want. The problem will be with the OTLP-profiling of the protocol. From the very beginning we should be ready to understand different sets of data. The most important bringing from OTEL world is [ebpf-profiling](https://github.com/open-telemetry/opentelemetry-ebpf-profiler). What is more, for sure the .NET data will be send in the OTel-compliant way. The data source is working, we need to add only the exporter.

Avoiding any Cisco/Splunk specific attributes should be our goal. If needed, we can start working on semantic-convention for profiling, but need to have good proposal for this.

tduncan:
I've mangled the commit signing and unable to figure out how to fix it. Instead I've opened #337 with correctly signed commits.

PR #337, opened 2025-02-19
Define profiling.instrumentation.source Log Message Attribute

tduncan:
The profiling backend ingestion process needs to be able to distinguish between callstacks collected by the existing continuous "Always On" profiler and those collected by the upcoming trace snapshot profiler. This PR defines an attribute to be included in a log message body that will identify which profiling process was responsible for collecting and reporting the exported callstacks.

See #336 for previous discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Semantic Conventions
+
+#### Enhancements
+
+- Define the `profiling.instrumentation.source` attribute with valid values
+  of `continuous` and `snapshot`.
+  [#336](https://github.com/signalfx/gdi-specification/pull/336)
+
 ## [1.7.0] - 2025-01-07
 
 ### Configuration
--- a/specification/semantic_conventions.md
+++ b/specification/semantic_conventions.md
@@ -129,6 +129,9 @@ instances. For each `LogRecord` instance:
   - `pprof-gzip-base64`,
   - `text` ([Deprecated](../README.md#versioning-and-status-of-the-specification)
     format).
+- `profiling.instrumentation.source` MUST be set to either:
+  - `continuous` for continuous "Always On" profiling
+  - `snapshot` for trace snapshot profiling
 
 #### `LogRecord` Message `text` Data Format Specific Attributes
 


Kielek (changes requested):
Needed to handle 2 things:
1. What if lack of the attribute: https://github.com/signalfx/gdi-specification/pull/336#discussion_r1959112445
2. Future consideration about OTLP protocol: https://github.com/signalfx/gdi-specification/pull/336#issuecomment-2667634684

tduncan:
Quoting from conversation started in https://github.com/signalfx/gdi-specification/pull/336#issuecomment-2667634684
>>>I am not against change for a short term solution, but we need to start thinking how to handle the OTLP profiling >>>protocol. I hope it will go live this year.
>>>In such cases you will be not able to expect such attributes. For known, libraries, you can probably utilize

>>>[InstrumentationScope](https://github.com/open-telemetry/opentelemetry->>>proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L138)
>>>For all generic cases:
>>>[SampleType](https://github.com/open-telemetry/opentelemetry->>>proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L192)
>>>[PeriodType](https://github.com/open-telemetry/opentelemetry->>>proto/blob/v1.5.0/opentelemetry/proto/profiles/v1development/profiles.proto#L225)
>>SampleType and PeriodType would seem to be describing the data itself, where as what we need to know is information >>about the context from which the callstacks were collected as the values for both SampleType and PeriodType would be >>identical. The shape of the profiling data does not vary between the existing continuous profiling solution and the trace >>snapshot extension we're porting into the Java agent.

>>I was hoping to use InstrumentationScope but conversations with individuals more familiar with the GDI process made it >>clear that, at least for now, the InstrumentationScope information will not be available for us to rely on.
>
>For our internal internal profiling format, we can put any data we want. The problem will be with the OTLP-profiling of the >protocol. From the very beginning we should be ready to understand different sets of data. The most important bringing >from OTEL world is [ebpf-profiling](https://github.com/open-telemetry/opentelemetry-ebpf-profiler). What is more, for sure >the .NET data will be send in the OTel-compliant way. The data source is working, we need to add only the exporter.
>
>Avoiding any Cisco/Splunk specific attributes should be our goal. If needed, we can start working on semantic-convention >for profiling, but need to have good proposal for this.

@Kielek, I agree and ideally we'd use `InstrumentationScope` but my understanding is our GDI pipeline is the limiting factor here, not the agent or backend. Can you think of another viable option for differentiating the source of the profiling stack traces?

laurit:
> Needed to handle 2 things:
> 
> 1. What if lack of the attribute: [Define `profiling.instrumentation.source` Log Message Attribute #336 (comment)](https://github.com/signalfx/gdi-specification/pull/336#discussion_r1959112445)

seems to be addressed

> 2. Future consideration about OTLP protocol: [Define `profiling.instrumentation.source` Log Message Attribute #336 (comment)](https://github.com/signalfx/gdi-specification/pull/336#issuecomment-2667634684)

I think we should deal with this when we have an otlp profiling ingest and an implementation that is able to send data to it. To me the most natural way seems to be using the sample type https://github.com/open-telemetry/opentelemetry-proto/blob/2bd940b2b77c1ab57c27166af21384906da7bb2b/opentelemetry/proto/profiles/v1development/profiles.proto#L192 If that doesn't work could also set an attribute on profile.

laurit:
> Sample type seems to be an awkward fit here. The comments for that field suggest its intended for the type of profiling data (cpu vs memory) with an associated unit. We'd need an additional dimension to the `ValueType` message since `continuous` profiles supports both cpu and memory.

The comment says 
```
  // For a cpu profile this might be:
  //   [["cpu","nanoseconds"]] or [["wall","seconds"]] or [["syscall","count"]]
```
I presume that at some point specification will define a set of well know values that represent wall time cpu profile, cpu time profile etc. Compliant backends can make sense of these samples and so something with them. Custom profilers, like our spnapshot profiler, that aren't specified in otel will use custom values and backends that don't recognize it can drop the samples.

Kielek (approved):

lachmatt:
Handled by https://github.com/signalfx/gdi-specification/pull/362

PR #338, opened 2025-03-19
AppD context -- First span is all we need.

breedx-splk:
Several people have brought up the point that putting the extra attributes on ALL spans with in a process shouldn't be necessary to paint the service map. Just the first one should suffice...so I'm hearing that feedback and applying it in this PR.


pellared (approved):
:+1:

seemk (approved):

PR #343, opened 2025-06-06
[DRAFT] Describe Trace Snapshot Profiling

tduncan:
First (rough) draft for defining how agents should profile traces selected for snapshotting. I'm very unfamiliar with writing this type of documentation so any help steering it where it needs to go is greatly appreciated!

### DIFF
--- a/specification/behaviors.md
+++ b/specification/behaviors.md
@@ -65,7 +65,7 @@ Agents MUST specify set the `profiling.instrumentation.source` value to `snapsho
 
 ### Starting Trace Profiler
 
-The OpenTelemetry Baggage entry for `splunk.snapshot.volume` MUST be used to 
+The OpenTelemetry Baggage entry for `splunk.trace.snapshot.volume` MUST be used to 
 decide whether to profile a trace. A value of `higest` is the signal to begin 
 profiling where as a value of `off` is an explicit signal to not profile.
 


specification/behaviors.md
```diff
@@ -51,3 +51,69 @@ metrics).
...
+
+Agents SHOULD take an initial stack trace sample when starting to profile a trace.
+
+### Trace Profiling
+An instrumentation library that has trace snapshot profiling capabilities MUST
+be able to sample call stacks for specific trace ids at a fixed interval.
+
+When a language runtime supports threading, stacks MUST be sampled only for 
+trace ids selected for snapshotting. The samples for profiled threads SHOULD be 
+taken instantaneously and MAY be taken at separate times.
```
laurit, line 87: > The samples for profiled threads SHOULD be taken instantaneously and MAY be taken at separate times.

I don't follow the meaning of this sentence. Also taking a stack trace is not an instantaneous operation by any means. In java for example stack trace is taken at a safepoint, which means that all threads are suspended (apparently recent vms are able to suspend only a single thread, but I don't know whether it is used when taking a stack trace).

specification/behaviors.md
```diff
@@ -51,3 +51,69 @@ metrics).
...
+
+Agents SHOULD attempt to export any remaining stack traces during the Agent shutdown phase. 
+
+### Call Stack Ingest
+
+Call stacks MUST be ingested as [OpenTelemetry
+Logs](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs).
+The logs containing profiling data MUST be sent via OTLP. Instrumentation
+libraries SHOULD reuse persistent OTLP connections from other signals (traces,
+metrics).
```
laurit, line 119: Although it is copied from https://github.com/signalfx/gdi-specification/blob/main/specification/behaviors.md#call-stack-ingest wanted to point out that I suspect that this is not true for the java implementation.

specification/semantic_conventions.md
```diff
@@ -272,3 +272,20 @@ For each `cpu` sample:
   in milliseconds if this sample represents a periodic event
 - label `thread.state` of type `string` OPTIONALLY can be set to describe
   the state of the thread
+
+## Trace Snapshot Profiling
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+Unless stated otherwise Agents MUST follow the `Profiling `ResourceLogs` Message`
```
laurit, line 280: > `Profiling `ResourceLogs` Message`

is this a typo?

specification/semantic_conventions.md
```diff
@@ -272,3 +272,20 @@ For each `cpu` sample:
...
+
+Trace Snapshot Profiling Configuration Options
+
+| Name                                         | Type   | Description                                         | Valid Values                 | Default |
+|----------------------------------------------|--------|-----------------------------------------------------|------------------------------| ------- |
+| `splunk.snapshot.profiler.enabled`           | string | Enable or Disable trace snapshot profiling          | `true` or `false`            | `false` |
+| `splunk.snapshot.profiler.sampling.interval` | string | Interval in which to take trace stack trace samples | Any valid duration `string`  | `10ms`  |
+
+The span attribute `splunk.snapshot.profiling` with a value of `true` indicates that 
+a trace within a service has been profiled.
```
laurit, line 291: maybe point out that the attribute should be se on the local root span?

specification/behaviors.md
```diff
@@ -51,3 +51,69 @@ metrics).
...
+This section describes the behavior for Splunk instrumentation libraries 
+that contain trace snapshot profiling features.
+
+### Instrumentation Source
+
+Agents MUST specify set the `profiling.instrumentation.source` value to `snapshot`
+
+### Starting Trace Profiler
+
+The OpenTelemetry Baggage entry for `splunk.snapshot.volume` MUST be used to
```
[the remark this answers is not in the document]
tduncan: Typo. Updated

Kielek:
Handled by #362

PR #344, opened 2025-06-10
Remove real user monitoring / rum

breedx-splk:
This is no longer maintained within GDI.

### DIFF
--- a/.github/ISSUE_TEMPLATE/ga_request.md
+++ b/.github/ISSUE_TEMPLATE/ga_request.md
@@ -51,11 +51,6 @@ Name and link to repository
     - [ ] Documents how to configure manual instrumentation.
     - [ ] Documents how to configure log correlation.
     - [ ] Documents minimum supported version of each auto-instrumentation framework.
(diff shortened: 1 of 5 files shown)


pjanotti (approved):

aditi-s3:
@breedx-splk Should https://github.com/signalfx/gdi-specification/blob/main/repositories.txt be updated to remove the RUM repositories then? So no spec changes does not trigger workflows in these repos?

laurit (approved):
> This is no longer maintained within GDI.

I think this isn't really important. The purpose of this document is to keep implementations aligned. If the RUM team wishes to host this document elsewhere then that is fine, though in my opinion they could keep using this repo too.

breedx-splk:
> @breedx-splk Should https://github.com/signalfx/gdi-specification/blob/main/repositories.txt be updated to remove the RUM repositories then? So no spec changes does not trigger workflows in these repos?

Yes, agreed, good point. I'll update it here.

pellared (approved):

PR #347, opened 2025-06-23
How to contribute to documentation

Kielek:


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 
 - `SPLUNK_PROFILER_CALL_STACK_INTERVAL` defaults to `1000` for single-threaded runtimes.
   [#325](https://github.com/signalfx/gdi-specification/pull/325)
+- Fix information how to contribute to documentation in
+  CONTRIBUTING template.
```
pellared, line 14:
```suggestion
  `CONTRIBUTING.md` template.
```

pellared (approved):

pellared:
> It's not really for _this_ project tho, is it? Why mention that here?

Thanks to this PR I know that I need to update the `CONTRIBUTING.md` in other repos  and we can double-check with @pkopta-splunk.

It can be also handy if we create a new repo.

laurit (approved):

breedx-splk:
> @breedx-splk, the old way to edit documentation (unfortunately) was changed. Previously there were a possibility to create PRs/Issues by any GitHub users. Now, there is a possibility only to create tickets for the documentation team. We should reflect it in all our repositories in the common way.

Oh! Thanks for clarifying. I completely missed that the fix-it-yourself ability had been removed. That sucks. Ok, this is a good idea. Thanks.

PR #355, opened 2025-11-11
Splunk specific file based configuration fields

seemk:
The following is a proposed addition to OTel's file based configuration. I've listed out our common configuration fields and how it should be placed into the OTel's configuration file.

Please discuss the wording and add any missing fields I've left out. Plus language maintainers if you have any distro specific configuration values, please add them here.

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -309,7 +309,7 @@ where it makes the most sense convention and usability wise.
 
 #### [Java SystemProperties](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html)
 
-These properties MUST match the environment variables converting to lower
+These properties MUST match the environment variables converted to lower
 case and replacing underscores with hyphens or periods. For example:
 system property `splunk.trace-response-header.enabled` is equivalent to environment
 variable `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED`.


specification/configuration.md
```diff
@@ -250,12 +250,51 @@ are required.
...
+
+In addition, Splunk specific configuration MUST have their own root-level configuration block named `splunk`.
+
+The following is an example covering the common configuration values:
+
+```yaml
+splunk:
+  # Common values across language distributions.
+  realm: "eu0"
+  accessToken: "${SPLUNK_ACCESS_TOKEN}"
```
Kielek: We have discussed it already. This should be configurable by standard OTel properties. Need to prepare only snippets.

specification/configuration.md
```diff
@@ -250,12 +250,51 @@ are required.
...
+    enabled: true # SPLUNK_SNAPSHOT_PROFILER_ENABLED
+    samplingInterval: 10 # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
+    selectionProbability: 0.01 # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
+  # Language specific configurations
+  runtime:
+    nodejs:
+      metrics:
+        instrumentationMetricsEnabled: true # SPLUNK_INSTRUMENTATION_METRICS_ENABLED
+      runtimeMetrics:
+        enabled: true # SPLUNK_RUNTIME_METRICS_ENABLED
```
Kielek: Same as in previous `enabled` property

specification/configuration.md
```diff
@@ -250,12 +250,51 @@ are required.
...
+#### File based configuration
+
+OpenTelemetry's [declarative configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/README.md#declarative-configuration) SHOULD be supported via [`OTEL_EXPERIMENTAL_CONFIG_FILE`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#via-otel_experimental_config_file) environment variable.
+
+In addition, Splunk specific configuration MUST have their own root-level configuration block named `splunk`.
+
+The following is an example covering the common configuration values:
+
+```yaml
+splunk:
```
robsunday: We may consider to name it splunk/development to indicate that it is not stable yet. This is [Otel convention](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/README.md#experimental-features), but I'm not sure if we really need it.

specification/configuration.md
```diff
@@ -250,12 +250,51 @@ are required.
...
+    memoryProfilingEnabled: true # SPLUNK_PROFILER_MEMORY_ENABLED
+  callgraphs:
+    enabled: true # SPLUNK_SNAPSHOT_PROFILER_ENABLED
+    samplingInterval: 10 # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
+    selectionProbability: 0.01 # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
+  # Language specific configurations
+  runtime:
+    nodejs:
+      metrics:
+        instrumentationMetricsEnabled: true # SPLUNK_INSTRUMENTATION_METRICS_ENABLED
```
robsunday: Why this kind of properties should not go under `.instrumentation.nodejs` ?

ysolomchenko: I agree that things related to instrumentations can be placed under the `.instrumentation.nodejs` node,
but the `profiler` `call graphs` and `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED ` should be made into a separate node `.splunk` or `.vendors.splunk`

specification/configuration.md
seemk: If the file is intended to be edited manually, then it's subpar UX. You'd have to modify all of the signal exporters, as a an end user I'd be asking "Couldn't you just make it editable in one location?".

For us it's definitely simpler like that.

specification/configuration.md
```diff
@@ -250,12 +250,51 @@ are required.
...
+
+The following is an example covering the common configuration values:
+
+```yaml
+splunk:
+  # Common values across language distributions.
+  realm: "eu0"
+  accessToken: "${SPLUNK_ACCESS_TOKEN}"
+  tracing:
+    enabled: true
```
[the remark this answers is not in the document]
Kielek: I agree, but we should avoid any snowflakes in our distributions as long as possible.
This discussion should go to the Config SIG level, not here.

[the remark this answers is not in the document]
seemk: Upstream is using the `disabled` key (which I changed to now), can you elaborate?

specification/configuration.md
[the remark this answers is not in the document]
seemk: Moved it to `distribution.splunk`

specification/configuration.md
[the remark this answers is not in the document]
seemk: Changed it to be disabled, the configuration is already pretty hostile to users (with the exporter header, endpoint duplication), I'm not really keen on removing a simple field to toggle the whole signal.

If we're going the route of removing `disabled` / `enabled` field, then:
* tracing / metrics / logs would be toggled via the existence of `tracer_provider` / `meter_provider` / `logger_provider` top level nodes
* profiling would be toggled via the existence of `distribution.splunk.profiling` node
* callgraphs would be toggled via the existence of `distribution.splunk.callgraphs` node

However as a compromise the `disabled` field under signals could be optional, where an unspecified value means the signal is on, so you'd explicitly have to write `disabled: true` in order to disable a signal without removing the whole config block (i.e. when you're debugging something).

specification/configuration.md
```diff
@@ -250,12 +250,73 @@ are required.
...
+        metrics:
+          runtime_metrics:
+            enabled: true # SPLUNK_RUNTIME_METRICS_ENABLED
+            collection_interval: 30000 # SPLUNK_RUNTIME_METRICS_COLLECTION_INTERVAL
+          instrumentation_metrics_enabled: true # SPLUNK_INSTRUMENTATION_METRICS_ENABLED
+          autoinstrument_packages:
+            - "MyApiService"
+            - "MyOtherService"
+        java:
+          ...
```
pellared: I suggest moving under `instrumentation` node.

specification/configuration.md
```diff
@@ -250,12 +250,73 @@ are required.
...
+      disabled: false
+      response_header_enabled: true # SPLUNK_TRACE_RESPONSE_HEADER_ENABLED
+    profiling:
+      disabled: false # SPLUNK_PROFILER_ENABLED
+      endpoint: "" # SPLUNK_PROFILER_LOGS_ENDPOINT
+      memory_profiling_enabled: true # SPLUNK_PROFILER_MEMORY_ENABLED
+    callgraphs:
+      disabled: false # SPLUNK_SNAPSHOT_PROFILER_ENABLED
+      sampling_interval: 10 # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
+      selection_probability: 0.01 # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
```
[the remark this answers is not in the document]
pellared: Note that the proposal misses some of the configuration supported by Java.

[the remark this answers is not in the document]
seemk: Giving it some thought again after our discussion, I think it doesn't make sense because of 2 reasons:
* Distros might not use the OpenTelemetry pipeline for profiling data exports. Node.js doesn't - we recently had a custom exporter (without using the OTel logs processors), but switched to an OTel logs exporter now. We are not using the logs processing pipeline and don't think we ever will.
* While the naming might be more correct, differentiating this from our current customer terminology and documentation is bound to sow confusion.

An unrelated to your comment, but a related issue is that while going over the configuration, I noticed that you can't actually set the endpoint for callgraphs explicitly with file based config. The [documentation](https://help.splunk.com/en/splunk-observability-cloud/monitor-application-performance/call-graph-profiling/get-data-into-call-graph-profiling) we have refers to `OTEL_EXPORTER_OTLP_ENDPOINT`. However this does not make sense much with file based config when the pipelines are in their own sections.

We either have to create a new configuration option for a callgraphs endpoint or we use the one from profiler.
Now both the CPU profiler and callgraphs can be active irrespective of each other. When (as previously proposed) we don't have an `enabled/disabled` flag under profiling we can end up in a case where we can't set the endpoint for callgraphs, e.g.:

```yaml
distribution:
  splunk:
    #profiling:
    #  endpoint: "collector:4318/v1/logs"
    callgraphs:
      trace_selection_probability: 0.05
      # Since profiling is commented out, we have no URL to use

#logger_provider:
#  processors:
#    - batch:
#        exporter:
#           otlp_http:
#              endpoint: ""  
```

Some of the options we have:
```yaml
distribution:
  splunk:
    profiling: # Signal is not enabled/disabled by the existence of the node
      disabled: true
      endpoint: "collector:4318/v1/logs"
    callgraphs:
      trace_selection_probability: 0.05
```
or
```yaml
distribution:
  splunk:
    # profiling: # Profiling is disabled
    #  endpoint: "collector:4318/v1/logs"
    callgraphs:
      endpoint: "collector:4318/v1/logs"
      trace_selection_probability: 0.05
```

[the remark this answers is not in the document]
ysolomchenko: If you don’t use the OTEL pipeline, then it can be like this.

```yaml
distribution:
  splunk:
    profiling:
      schedule_delay: 5000 # SPLUNK_PROFILER_EXPORT_INTERVAL
      export_timeout: 30000 # SPLUNK_PROFILER_EXPORT_TIMEOUT
      exporter:
        otlp_log_http: # SPLUNK_PROFILER_OTLP_PROTOCOL for java can be otlp_log_grpc
          endpoint: http://localhost:4318/v1/logs # SPLUNK_PROFILER_LOGS_ENDPOINT
      memory_allocation_sampling: # SPLUNK_PROFILER_MEMORY_ENABLED
        sampling_interval / sampling_rate: # SPLUNK_PROFILER_MEMORY_EVENT_RATE /  SPLUNK_PROFILER_MAX_MEMORY_SAMPLES ?
      callstack_sampling: # SPLUNK_PROFILER_ENABLED
        sampling_interval: 10  # SPLUNK_PROFILER_CALL_STACK_INTERVAL
      snapshoting: # SPLUNK_SNAPSHOT_PROFILER_ENABLED
        callstack_sampling_interval: 10  # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
        trace_selection_probability: 0.01 # SPLUNK_SNAPSHOT_SELECTION_PROBABILTY
```

or 

``` yaml
distribution:
  splunk:
    profiling:
      exporter:
        otlp_log_http: # SPLUNK_PROFILER_OTLP_PROTOCOL for java can be otlp_log_grpc
          schedule_delay: 5000 # SPLUNK_PROFILER_EXPORT_INTERVAL
          export_timeout: 30000 # SPLUNK_PROFILER_EXPORT_TIMEOUT
          endpoint: http://localhost:4318/v1/logs # SPLUNK_PROFILER_LOGS_ENDPOINT
      memory_allocation_sampling: # SPLUNK_PROFILER_MEMORY_ENABLED
        sampling_interval / sampling_rate: # SPLUNK_PROFILER_MEMORY_EVENT_RATE /  SPLUNK_PROFILER_MAX_MEMORY_SAMPLES ?
      callstack_sampling: # SPLUNK_PROFILER_ENABLED
        sampling_interval: 10  # SPLUNK_PROFILER_CALL_STACK_INTERVAL
      snapshoting: # SPLUNK_SNAPSHOT_PROFILER_ENABLED
        callstack_sampling_interval: 10  # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
        trace_selection_probability: 0.01 # SPLUNK_SNAPSHOT_SELECTION_PROBABILTY
```

I also think we don’t need multiple processors for profiling.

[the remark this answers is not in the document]
pellared: > I think we should use the term from the spec, otlp_http/otlp_grpc.

The problem is that we are using a "log OTLP exporter" not "profiling OTLP expoter" which we may want to in future. This is why I added `_log_`  :wink:

specification/configuration.md
```diff
@@ -250,12 +250,79 @@ are required.
...
+    profiling:
+      exporter:
+        otlp_http:
+          endpoint: ""              # SPLUNK_PROFILER_LOGS_ENDPOINT
+      cpu_profiler:                 # SPLUNK_PROFILER_ENABLED
+        sampling_interval: 10       # SPLUNK_PROFILER_CALL_STACK_INTERVAL
+      memory_profiler:              # SPLUNK_PROFILER_MEMORY_ENABLED
+      callgraphs:                   # SPLUNK_SNAPSHOT_PROFILER_ENABLED
+        sampling_interval: 10       # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
+        selection_probability: 0.01 # SPLUNK_SNAPSHOT_SELECTION_PROBABILITY
```
robsunday: I'd like to suggest one change in profiling settings.
It is related to the fact that CPU profiling and memory profiling are described as a part of AlwaysOn Profiling. Also in Java distribution both types of profiling are based on the same mechanism: JFR.
There is set of 6 additional settings for JFR that are specific for Java and they are shared between CPU and memory profiling. Below is example of the profiling configuration with all the additional settings:
```yaml
distribution:
  splunk:
    profiling:
      exporter:
        otlp_http:                  # SPLUNK_PROFILER_OTLP_PROTOCOL
          endpoint: ""              # SPLUNK_PROFILER_LOGS_ENDPOINT

      always_on:
        include_agent_internals: true # SPLUNK_PROFILER_INCLUDE_AGENT_INTERNALS (Java)
        include_jvm_internals: true   # SPLUNK_PROFILER_INCLUDE_JVM_INTERNALS   (Java)
        stack_depth_limit: 256        # SPLUNK_PROFILER_MAX_STACK_DEPTH         (Java)
        keep_recording_files: true    # SPLUNK_PROFILER_KEEP_FILES              (Java)
        recording_directory: "/tmp/profiler" # SPLUNK_PROFILER_DIRECTORY        (Java)
        recording_duration: 20000     # SPLUNK_PROFILER_RECORDING_DURATION      (Java)

        cpu_profiler:               # SPLUNK_PROFILER_ENABLED
          sampling_interval: 10     # SPLUNK_PROFILER_CALL_STACK_INTERVAL
        memory_profiler:            # SPLUNK_PROFILER_MEMORY_ENABLED
          event_rate:               # SPLUNK_PROFILER_MEMORY_EVENT_RATE         (Java)
          event_rate_limit:         # SPLUNK_PROFILER_MEMORY_EVENT_RATE_LIMIT   (Java)
          native_sampling:          # SPLUNK_PROFILER_MEMORY_NATIVE_SAMPLING    (Java)

      call_graphs:                  # SPLUNK_SNAPSHOT_PROFILER_ENABLED
        sampling_interval: 10       # SPLUNK_SNAPSHOT_SAMPLING_INTERVAL
        export_interval: 5000       # SPLUNK_PROFILER_EXPORT_INTERVAL           (Java)
        selection_probability: 0.01 # SPLUNK_SNAPSHOT_SELECTION_PROBABILITY
        staging_capacity: 1000      # SPLUNK_SNAPSHOT_PROFILER_STAGING_CAPACITY (Java)
        stack_depth_limit: 512      # SPLUNK_SNAPSHOT_PROFILER_MAX_STACK_DEPTH  (Java)
```

I grouped cpu_profiler and memory_profiler nodes under common parent: always_on. All common JFR properties specific to Java are also under this node. All of them are optional, and should be ignored by non Java distros.
What do you think about this idea?

robsunday:
I also think that this stuff should go under `.instrumentation/development.{lang}`. It is related to the instrumentation, as @ysolomchenko pointed out.
Additionally there is a very small number of options common for all languages.
`.instrumentation/development` is already designed with multiple languages in mind, while `.distribution` is not, and we would have to mimic the structure of `.instrumentation/development` to avoid a mess

> Keeps Splunk specific configuration localized instead of having it all over the place in the config file. Meaning everything outside of distribution.splunk would be standard OTel config.

Some parts of Splunk specific configuration is already spread across the config file, because every custom `ComponentProvider` can be referenced only from a dedicated YAML location.

seemk:
Anyway I'm fine with going the now proposed structure.

specification/configuration.md
```diff
@@ -239,24 +239,99 @@ are required.
...
+        collection_interval: 30000
+    traces:
+      http:
+        splunk_trace_response_header_enabled: true # SPLUNK_TRACE_RESPONSE_HEADER_ENABLED
+        splunk_capture_uri_parameters:
+          - "userId"
+      redis:
+        splunk_include_command_args: true    # SPLUNK_REDIS_INCLUDE_COMMAND_ARGS
+      pg:
+        disabled: true
```
ysolomchenko: As we already have `splunk:` as a semi-node here, we can remove the `splunk_` prefix from the properties.

seemk: The 2 issues I have with this:
1) Duplicated configuration option (aspnet, aspnetcore), correct me if I'm wrong about .net, but on Node both programmatic and env var based config has the option to toggle it in a single place.
2) As an user, is it intuitive that you need to know the instrumentation names for our distro specific config? One point of our distros has been to remove the gnarlyness of OTel behind sensible, easy options

ysolomchenko (approved):

specification/configuration.md
```diff
@@ -239,24 +239,80 @@ are required.
...
+    profiling:
+      exporter:
+        otlp_http:
+          endpoint: ""                # SPLUNK_PROFILER_LOGS_ENDPOINT
+      # always_on, cpu_profiler, memory_profiler, callgraphs may contain language specific key value pairs.
+      always_on:
+        cpu_profiler:                 # SPLUNK_PROFILER_ENABLED
+          sampling_interval: 10       # SPLUNK_PROFILER_CALL_STACK_INTERVAL
+        memory_profiler:              # SPLUNK_PROFILER_MEMORY_ENABLED
+      callgraphs:                     # SPLUNK_SNAPSHOT_PROFILER_ENABLED
```
ysolomchenko:
```suggestion
      always_on:
        cpu_profiler:                 # SPLUNK_PROFILER_ENABLED
          sampling_interval: 10       # SPLUNK_PROFILER_CALL_STACK_INTERVAL
        memory_profiler:              # SPLUNK_PROFILER_MEMORY_ENABLED
          sampling_interval: 10
      callgraphs:                     # SPLUNK_SNAPSHOT_PROFILER_ENABLED
```

specification/configuration.md
```diff
@@ -239,24 +239,80 @@ are required.
...
+configuration block named `splunk`.
+
+The following is a configuration blueprint covering the common values:
+
+```yaml
+distribution:
+  splunk:
+    profiling:
+      exporter:
+        otlp_http:
```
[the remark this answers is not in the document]
ysolomchenko: https://github.com/signalfx/gdi-specification/pull/355#discussion_r2567852519

specification/configuration.md
robsunday: > In fact, I think the lang distros can put whatever they want in their lang block. We're spending way too much time on trivial configuration fields.

I think it provides good solution for this.

specification/configuration.md
robsunday: Yes, Java use a rate, but for consistency I'll use sampling_interval and do necessary calculations/transformations at runtime

seemk:
> Quick question about `SPLUNK_ACCESS_TOKEN` and `SPLUNK_REALM`. Are we blocking them (no usage), or should we use environment variables with a file-based approach?

Env vars with a file based approach would be the recommended then, in case you don't want the tokens in the file

PR #358, opened 2025-11-20
Prevent zero probability for snapshots

breedx-splk:
A zero value would be (should be!) semantically equivalent to never performing a snapshot, which can be achieved by just not enabling the feature in the first place.

Allowing a zero-probability just opens up room for confusion. For instance, in java, we currently allow it via system property/env var and when we do the random probabilistic determination there is (in the zero case) still a small chance that a trace will be sampled/selected. This aims to fix that consistently across languages by disallowing the zero probability case in the first place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
   validate-documentation:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5.0.1
+      - uses: actions/checkout@v6.0.0
       - name: Markdown Lint check
         uses: DavidAnson/markdownlint-cli2-action@v21.0.0
         with:
--- a/.github/workflows/lychee-weekly.yml
+++ b/.github/workflows/lychee-weekly.yml
@@ -5,7 +5,7 @@ jobs:
   linkChecker:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5.0.1
+      - uses: actions/checkout@v6.0.0
       - name: Link Checker
         uses: lycheeverse/lychee-action@v2.7.0
         with:


pellared (changes requested):
This is a breaking change. I am worried that someone already uses `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY=0` to disable snapshots.

breedx-splk:
> This is a breaking change. I am worried that someone already uses `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY=0` to disable snapshots.

I seriously doubt that. This is a new feature. Regardless, if they're doing that in java, they're in for a surprise. Are you confident that the languages are doing this consistently?

Also, we haven't released any versions of this since the prior PR was merged.

Kielek (approved):

Kielek:
I am fine with this change, .NET is still in the beta phase, we can make this changes this/early next week.

pellared:
> > This is a breaking change. I am worried that someone already uses `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY=0` to disable snapshots.
> 
> I seriously doubt that. This is a new feature. Regardless, if they're doing that in java, they're in for a surprise. Are you confident that the languages are doing this consistently?
> 
> Also, we haven't released any versions of this since the prior PR was merged.

Shoudn't the `Snapshot Profiler` section be marked as `Development`? Right now, I assumed it is Stable. 

> Regardless, if they're doing that in java, they're in for a surprise. Are you confident that the languages are doing this consistently?

I have no idea. But, I want us to make sure that we won't break any users if possible.

seemk:
I'm ok with the change, but what would be the behaviour when 0 is set? Set it to the default value and warn?

PR #361, opened 2025-11-27
Change Experimental status to Development to follow OTel Specification

pellared:
Follows https://github.com/open-telemetry/opentelemetry-specification/pull/4061

It also fixes a bug that "Snapshot Profiler" section was Stable. This is not following

> Everything in the specification starts as Experimental.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Specification
+
+#### Enhancements
+
+- Change Experimental status to Development to follow the OpenTelemetry
+  Specification terminology.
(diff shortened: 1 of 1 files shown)


pellared:
> Did you have chance what wording is used on the Splunk documentation side? Changes looks good, but I am not sure about wording used in the documentation. With current docs-system it might be not trivial to adjust all pages...

Agreed. Let's keep our current Splunk vocabulary.

> It also fixes a bug that "Snapshot Profiler" section was Stable. This is not following

> > Everything in the specification starts as Experimental.

I will make a PR for it once I am back at work.

PR #364, opened 2025-12-08
Mark Snapshot Profiler configuration as Experimental

pellared:
Follows https://github.com/signalfx/gdi-specification/pull/362

So that we have a chance to double-check when stabilizing the feature.

If this is indeed stable then changes like https://github.com/signalfx/gdi-specification/pull/358 should not be allowed. If this is a bugfix then it should be at least mentioned in the changelog as a bugfix. However, I think it is questionable if this was a bug. Note that e.g. `parentbased_traceidratio` accepts `0` as a valid value (reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration).

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -329,7 +329,7 @@ variable configurations MUST be available:
 | `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY` | 0.01    | The probability of a trace being sampled. MUST be 0 < n <= 1  |
 
 - [1]: The default sampling interval is runtime-specific. Known values are:
-  - .NET: 20ms
+  - .NET: 40ms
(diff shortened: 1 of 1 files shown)


pellared:
@breedx-splk, @seemk , PTAL

pellared:
> I think #358 should be reverted, zero probability makes sense. 🤔 Approved it too hastily

@seemk, do I understand correctly that you just prefer to revert #358 and keep this section Stable (as you have not approved this PR :wink: )?

seemk (approved):

Kielek (approved):

PR #366, opened 2025-12-11
Revert "Prevent zero probability for snapshots"

Kielek:
Reverts signalfx/gdi-specification#358

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -327,7 +327,7 @@ variable configurations MUST be available:
 | `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY` | 0.01    | The probability of a trace being sampled. MUST be 0 <= n <= 1 |
 
 - [1]: The default sampling interval is runtime-specific. Known values are:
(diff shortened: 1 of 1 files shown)


pellared:
It would be good to add the rationale in the PR description.

From https://github.com/signalfx/gdi-specification/pull/364:

> If this is indeed stable then changes like https://github.com/signalfx/gdi-specification/pull/358 should not be allowed. If this is a bugfix then it should be at least mentioned in the changelog as a bugfix. However, I think it is questionable if this was a bug. Note that e.g. parentbased_traceidratio accepts 0 as a valid value (reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration).

> I think https://github.com/signalfx/gdi-specification/pull/358 should be reverted, zero probability makes sense. 🤔 Approved it too hastily

Kielek:
Closing based on internal conversation with @seemk and @breedx-splk.

PR #367, opened 2025-12-12
Modified the requirements for trace selection for snapshot profiling.

PeterF778:
The sampling decisions and snapshot profiling decisions should be independent.
Selection of trace for snapshot profiling can be done based on independent source of entropy.
If it is desired that the decision is based on trace-id, the algorithm can use a hash of trace-id, assuming a "good" hash function, which provides uniformly distributed values for traces-ids belonging to a subinterval of all available trace-id values.


PeterF778:
I have read the CLA Document and I hereby sign the CLA

PR #373, opened 2026-01-05
Add instructions on release notes

atoulme:


specification/repository.md
```diff
@@ -170,6 +170,10 @@ to the repo.
 - MUST update all examples in the [Splunk OpenTelemetry example
   repository](https://github.com/signalfx/tracing-examples/tree/main/opentelemetry-tracing)
   that depends on the repository to use the latest release.
+- MUST list any upstream changes associated with the release, reproducing
+  fragments of the upstream changelog as needed.
+  - MUST call out any breaking change associated with an upstream release,
+    adding and linking to documentation steps to address the change.
```
breedx-splk: I don't think that duplicating this information (by copying it from upstream release notes into our release notes) is helpful. In fact, it probably detracts from what's important about the local changes. Linking to the release notes should suffice.

Are there other examples of prior art, where open source projects just copy in the list of changes from their dependencies?

atoulme:
Reiterating that I need this and we continue to do this for the collector with success.

specification/repository.md
atoulme: Rebounding here on this discussion:
* Prior art exists as demonstrated - if you'd like more, we have our changelog for the collector here: https://github.com/signalfx/splunk-otel-collector/releases/tag/v0.150.0
* I am asking this so we have rich, instructive release notes that reduce the surprises that our customers can have on upgrade.
* The argument that we should make sure local changes are emphasized is moot because most of the changes we make are actually upstream and deserve a lot more visibility. If anything, we are not giving a great signal to customers that they should upgrade to get new functionalities.
  To pick an example, the release notes for the latest .NET release are here: https://github.com/signalfx/splunk-otel-dotnet/releases/tag/v1.14.0 and mention upstream changes, which itself links to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/tag/v1.15.0-beta.1 which links to https://github.com/open-telemetry/opentelemetry-dotnet/releases/tag/core-1.15.2 which itself links to a CHANGELOG file. This makes it difficult for support and customers to follow what is released and to recommend updates.

breedx-splk:
@signalfx/gdi-specification-maintainers please chime in.

specification/repository.md
pjanotti: On the case of the collector the important point is that we are actively filtering in/out upstream changes that apply to the users of our distribution. The upstream release notes are really long and the filtering is really valuable to our users.

While the proposed wording capture this correctly, it doesn't give the flexibility to deliverables that are on the whole subject to upstream changes just to refer to upstream notes. 

I think flexibility ideal here.

Kielek:
I would like to see some POC to automate for maintaining CHANGELOG/release notes.
It probably require adjustments across various repositories (at least in .NET word).

I am against merging this PR before we have some working example. If we do not implement it, we will be not able to mark that we met newer gdi-spec. It migth be required eg for OpAmp stuff.

pellared (changes requested):
This changes does not seem to make sense for https://github.com/signalfx/splunk-otel-go which is a library. I do not think that users of OTel Go distro would expect us to copy the changelog of the dependencies. Also nothing prevents users from using "older version" of our distro together with "newer version" of upstream.

Note also that we already have:

> MUST state version of OpenTelemetry repository built against if applicable


I remember that this was been extensively discussed somewhere outside of GitHub and this is was the consensus: https://github.com/signalfx/gdi-specification/pull/68

seemk:
I've been thinking of adding the upstream changes to our release notes, just so it would be easier to see what changed. This can be helpful for other developers who are not doing the release, as usually the one doing the upstream upgrade needs to skim through the changes anyway, could very well just copy them to our release notes for clarity.

For the Node.js distro it would work great, as our upstream dependencies are specified by exact version.

`MUST list any upstream changes associated with the release` - I'm understanding it that we are allowed to filter out relevant changes from the upstream (as we don't use all the upstream packages). This would work great.

seemk (approved):

PR #379, opened 2026-03-03
Introduce opamp data model for agents, starting with identity

breedx-splk:
We need to flesh out some common data model for agents that wish to speak opamp to a server. This is the start of that. It is based on the upstream opamp spec and some additional requirements from the agent management team.

We will be following this up with additional data model and formats, like around reporting and receiving config.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -8,7 +8,7 @@ an OpAMP server. For more information about OpAMP, please see the
 
 ## Identifying Attributes
 
-The `AgentDescription` message includes a set of 
+The `AgentDescription` message includes a set of
 [identifying attributes](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes).
 Agents MUST copy these attributes from the OpenTelemetry resource and
 MUST use existing semantic conventions when they exist. Any attribute
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -1,0 +1,42 @@
+# OpAMP Data Model
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+This document describes agent data model fields when configured to speak with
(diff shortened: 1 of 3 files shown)


specification/opamp_datamodel.md
```diff
@@ -0,0 +1,42 @@
+# OpAMP Data Model
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+This document describes agent data model fields when configured to speak with
```
Kielek: Why not pass all available resource attributes? It can be up to the server how to handle all others?

breedx-splk: It's odd to me that the set of attributes is framed as "it's what the collector uses to self-identify in its own metrics". To agents, a resource is just a resource is THE resource, and really, aside from warning when `service.name` isn't present, we don't care how many or which resource attributes are detected and included.

Any resource attribute that we can find or are configured for, will be included (including the example `os.version`). I do get that the collector is a special case there, but it's odd that the OpAMP spec was(is?) so tied to that...but I get it, a bunch of the original intent was around managing and re/configuring collectors.

> I think it is OK to apply the same logic to language SDKs and include all Resource attributes in the identifying attributes.

I think it's fine as well. It's potentially wasteful and more resource intensive, but let's assume for now that it can be filtered and partitioned into whatever the server/backend needs are. I'll rewrite this PR then. 👍🏻

breedx-splk:
> @breedx-splk can you make it clear that `specification/opamp_datamodel.md` applies to language agents only? I see some confusion from people because we read OpAMP spec section which is about Otel Collector and try to apply that section to language agents, but that's not what OpAMP spec says.

Sure, I can add some clarification. 

It really is frustrating that people still refer to or think of the collector as an agent.

Kielek (approved):

seemk (approved):

breedx-splk:
Note: If https://github.com/open-telemetry/opamp-spec/pull/312 gets merged we will roll this back and do what upstream advises.

PR #385, opened 2026-04-08
Require use of GHA hash instead of version tag/string

breedx-splk:
GHA version tags/strings are immutable. This is very problematic and leaves opportunity for attackers to retroactively modify existing usages of actions.

I will not be able to make all these changes myself, but it is pretty straightforward to automate using contemporary tooling. Some examples:

* https://github.com/signalfx/splunk-otel-js/pull/1106
* https://github.com/signalfx/splunk-otel-java/pull/2740

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 
 #### Enhancements
 
+- Require referencing GitHub Actions by commit hash.
+  [#385](https://github.com/signalfx/gdi-specification/pull/385)
+
 #### Bugfixes
 
 ### Repository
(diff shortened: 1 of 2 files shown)


pjanotti (approved):

Kielek (approved):
Please add entry to the CHANGELOG.

I think that it can be extended for other technologies. Dockers images, python, package locks in js, etc. Everywhere you can pin, you should.
On the OTel side, we scan it by https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/fd7d486cb989654942204874d7a2d0e533142cdb/.github/workflows/ossf-scorecard.yml#L26

pellared (approved):
Please add changelog entry

pellared (commented):
An easy way to do it is via https://docs.renovatebot.com/upgrade-best-practices/

laurit (approved):

seemk (approved):

PR #390, opened 2026-05-14
Update declarative config env var name

breedx-splk:
The old experimental var is deprecated. Let's update to match.


laurit (approved):

Kielek (approved):

Kielek:
The only missing part is CHANGELOG entry.

PR #391, opened 2026-05-14
Effective configuration specification

### LINKED ISSUE #380: Decide on Effective Config structure for language agents
We need to decide how to record config in `effectiveConfig` field.

Copying notes that @breedx-splk wrote:
```
Formats may be different, depending on how the agent was initially configured
  - env vars vs declarative config
  - Jason thinks we agreed to use different formats, but I think Rasmus (and maybe others?)
    have pushed back and prefer a single format in all cases.
  - Jason was thinking `key=value\n` when env and filtered yaml when declarative config.
```

breedx-splk:
Resolves #380.

Here is an initial spec for expressing declarative config through opamp. It has two sections: one for when the agent is configured with declarative config and one without. It specifies the filenames, content types, and body format/structure under those two conditions.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -99,18 +99,20 @@ Here is a description of this format in Backus-Naur Form:
 
 Note: <value-char> is any character that's not a newline
 
+The order of these lines is not important. Each <key> MUST only appear in the body once.
+
 #### Required fields:
 
 The following configuration items MUST be reported in the body:
 
+* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
+* `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
+* `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`
 * `SPLUNK_PROFILER_ENABLED`
 * `SPLUNK_PROFILER_MEMORY_ENABLED`
 * `SPLUNK_SNAPSHOT_PROFILER_ENABLED`
 * `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL`
 * `SPLUNK_PROFILER_CALL_STACK_INTERVAL`
-* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
-* `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
-* `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`
 
 Additional configuration items SHOULD NOT be provided by agents.
 
@@ -133,14 +135,14 @@ environment variables are intentionally omitted from this format.
 * content_type: `text/plain; format=properties; vendor=splunk; v=1`
 
 ```properties
-SPLUNK_PROFILER_ENABLED=true
-SPLUNK_PROFILER_MEMORY_ENABLED=false
-SPLUNK_SNAPSHOT_PROFILER_ENABLED=false
-SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL=0
-SPLUNK_PROFILER_CALL_STACK_INTERVAL=0
 OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
 OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://localhost:4318/v1/metrics
 OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=http://localhost:4318/v1/logs
+SPLUNK_PROFILER_ENABLED=true
+SPLUNK_PROFILER_MEMORY_ENABLED=false
+SPLUNK_PROFILER_CALL_STACK_INTERVAL=1001
+SPLUNK_SNAPSHOT_PROFILER_ENABLED=false
+SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL=0
 ```
 
 ### Effective Declarative Config
@@ -167,4 +169,74 @@ When reporting `EffectiveConfig`, the following MUST be followed:
 
 #### Body Format
 
-The body yaml format MUST match the structure
+The body yaml format MUST follow the structure of the
+[OpenTelemetry Declarative Configuration](https://opentelemetry.io/docs/languages/sdk-configuration/declarative-configuration/)
+specification, but will be a more minimal, filtered representation that contains only the
+configuration items that we have deemed important enough to report in "effective config".
+
+Some of these effective configuration values have a defined location in the yaml structure,
+while some are Splunk specific.
+
+#### Required fields:
+
+The following uses "dot path" shorthand for the equivalent yaml fields.
+The following configuration items SHOULD be reported in the body:
+
+* `tracer_provider.exporter.otlp_http.endpoint`
+* `meter_provider.readers[<n>].periodic.exporter.otlp_http.endpoint`
+* `logger_provider.exporter.otlp_http.endpoint`
+* `distribution.splunk.profiling.always_on.cpu_profiler.sampling_interval`
+* `distribution.splunk.profiling.always_on.memory_profiler`
+* `distribution.splunk.profiling.callgraphs.sampling_interval`
+
+Unlike the environment variable effective config, the existence of some values
+in the structure will imply that some features are enabled. For example, if
+the cpu profiler sampling interval is configured, this implies that profiling
+is enabled. Similarly, if this value is not present, it implies that the cpu profiler
+is not enabled.
+
+Because this is "effective config", default values MUST be provided even when they are
+absent in the physical yaml file on disk, unless providing a value were to change the 
+semantics of absence.
+
+The OTel declarative configuration specification allows environment variables to 
+be used inside strings via templates. When this is used, the final template evaluated
+value MUST be provided, and not the environment variable's name.
+
+Unlike the environment variable effective config, when multiple exporters are configured for a
+given signal, the agent SHOULD provide all the actively used endpoints.
+
+#### Example:
+
+```yaml
+tracer_provider:
+  processors:
+    - batch:
+        exporter:
+          otlp_http:
+            endpoint: http://localhost:4318/v1/traces
+  meter_provider:
+    readers:
+      - periodic:
+          exporter:
+            otlp_grpc:
+              endpoint: http://localhost:4318/v1/metrics
+  logger_provider:
+    processors:
+      - simple:
+          exporter:
+            otlp_http:
+              endpoint: http://localhost:4318/v1/logs
+distribution:
+  splunk:
+    profiling:
+      always_on:
+        cpu_profiler:
+          sampling_interval: 1001
+        memory_profiler:
+      callgraphs:
+        sampling_interval: 10
+```
+
+Note: The true configuration file may be significantly larger or more complicated than what is
+actually provided via effective configuration.
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -73,7 +73,7 @@ is assumed that its configuration is derived from environment variables
 and default values. This section only applies to agents that were
 started *without* declarative configuration.
 
-When reporting `EffectiveConfig`, the following MUST be followed:
+When reporting `EffectiveConfig`, the following SHOULD be followed:
 
 * The `ConfigMap` MUST contain an `AgentConfigFile` under the name
   `environment`.
@@ -84,6 +84,9 @@ When reporting `EffectiveConfig`, the following MUST be followed:
   `AgentConfigFile` `body`. The v field
   allows us to revise this format in a backwards compatible way in the future.
 * The `AgentConfigFile` body MUST conform to the body format below:
+
+Agents MAY choose to map environment configuration into the declarative 
+format described below.
 
 #### Body Format
 
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -206,6 +206,9 @@ value MUST be provided, and not the environment variable's name.
 Unlike the environment variable effective config, when multiple exporters are configured for a
 given signal, the agent SHOULD provide all the actively used endpoints.
 
+Implementations MAY choose to include disabled/inactive features in the yaml 
+document with commented out lines.
+
 #### Example:
 
 ```yaml
@@ -234,9 +237,11 @@ distribution:
         cpu_profiler:
           sampling_interval: 1001
         memory_profiler:
-      callgraphs:
-        sampling_interval: 10
+#      callgraphs:
+#        sampling_interval: 10
 ```
+
+Note: The absence of callgraphs sampling_interval here indicates that the feature is not active.
 
 Note: The true configuration file may be significantly larger or more complicated than what is
 actually provided via effective configuration.
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -125,13 +125,16 @@ The following configuration items MUST be reported in the body:
 * `SPLUNK_SNAPSHOT_PROFILER_ENABLED`
 * `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL`
 * `SPLUNK_PROFILER_CALL_STACK_INTERVAL`
+* `OTEL_CONFIG_FILE`
+* `OTEL_EXPERIMENTAL_CONFIG_FILE`
 
 Additional configuration items SHOULD NOT be provided by agents.
 
 If any of these configuration items is not specified via environment
 variables, then a default value MUST be provided. The value provided
 MUST be semantically equivalent to the currently used value, regardless
-of what was actually supplied in the environment.
+of what was actually supplied in the environment. String values that 
+do not have a default SHOULD provide `null`.
 
 For example, even if the environment contains `SPLUNK_PROFILER_ENABLED=true`,
 but the agent could not enable the profiler due to runtime platform limitations,
@@ -154,6 +157,8 @@ SPLUNK_PROFILER_MEMORY_ENABLED=false
 SPLUNK_PROFILER_CALL_STACK_INTERVAL=1001
 SPLUNK_SNAPSHOT_PROFILER_ENABLED=false
 SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL=0
+OTEL_CONFIG_FILE=null
+OTEL_EXPERIMENTAL_CONFIG_FILE=null
 ```
 
 ### Effective Declarative Config
@@ -204,6 +209,8 @@ The following configuration items SHOULD be reported in the body:
 * `distribution.splunk.profiling.always_on.cpu_profiler.sampling_interval`
 * `distribution.splunk.profiling.always_on.memory_profiler`
 * `distribution.splunk.profiling.callgraphs.sampling_interval`
+* `otel.config.file`
+* `otel.experimental.config.file`
 
 Unlike the environment variable effective config, the existence of some values
 in the structure will imply that some features are enabled. For example, if
@@ -230,6 +237,8 @@ document with commented out lines.
 #### Example
 
 ```yaml
+otel.config.file: /usr/otel/agent.yaml
+otel.experimental.config.file: null
 tracer_provider:
   processors:
     - batch:
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -76,8 +76,8 @@ started *without* declarative configuration.
 When reporting `EffectiveConfig`, the following MUST be followed:
 
 * The `ConfigMap` MUST contain an `AgentConfigFile` under the name
-  `environment_variables`.
-* The `AgentConfigFile` under `environment_variables` MUST have a
+  `environment`.
+* The `AgentConfigFile` under `environment` MUST have a
   `content_type` of
   `text/plain; format=properties; vendor=splunk; v=1.0.0`. This content type
   tells the remote side how to interpret the content within the
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -240,8 +240,8 @@ document with commented out lines.
 #### Example
 
 ```yaml
-otel.config.file: /usr/otel/agent.yaml
-otel.experimental.config.file: null
+otel_config_file: /usr/otel/agent.yaml
+otel_experimental_config_file: null
 tracer_provider:
   processors:
     - batch:
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -145,6 +145,11 @@ OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=http://localhost:4318/v1/logs
 
 ## Effective Declarative Config
 
-When an agent has been started _with_ declarative configuration, it is assumed that its
-configuration is derived from environment variables and default values. This section only applies
-to agents that were started _without_ declarative configuration.
+When an agent has been started _with_ a declarative configuration yaml file, its
+"effective configuration" will be reported differently. This section only applies to agents that
+were started _with_ declarative configuration.
+
+When running with declarative configuration, the "effective configuration" body will 
+follow a similar structure, but will be a subset of 
+
+### Body Format
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -60,7 +60,7 @@ In the spec below, for simplicity, we treat java system properties as environmen
 variables (they are indeed
 [interchangeable](https://opentelemetry.io/docs/zero-code/java/agent/configuration/#configuring-with-environment-variables)).
 
-## Effective Environment Config
+### Effective Environment Config
 
 When an agent has been started _without_ declarative configuration, it is assumed that its
 configuration is derived from environment variables and default values. This section only applies
@@ -75,7 +75,7 @@ When reporting `EffectiveConfig`, the following MUST be followed:
   allows us to revise this format in a backwards compatible way in the future.
 * The `AgentConfigFile` body MUST conform to the body format below:
 
-### Body Format
+#### Body Format
 
 The body MUST be a list of text lines, separated by newlines. Each line MUST
 be in the form `<key>=<value>\n` where the `<key>` is the name of the configuration
@@ -143,13 +143,28 @@ OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=http://localhost:4318/v1/metrics
 OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=http://localhost:4318/v1/logs
 ```
 
-## Effective Declarative Config
+### Effective Declarative Config
 
 When an agent has been started _with_ a declarative configuration yaml file, its
 "effective configuration" will be reported differently. This section only applies to agents that
 were started _with_ declarative configuration.
 
 When running with declarative configuration, the "effective configuration" body will 
-follow a similar structure, but will be a subset of 
+follow a similar structure, but will be a subset of what the actual file supports.
 
-### Body Format
+When reporting `EffectiveConfig`, the following MUST be followed:
+
+* The `ConfigMap` MUST contain an `AgentConfigFile` under the name that exactly matches the 
+  filename provided in the `OTEL_CONFIG_FILE` environment variable. This MUST include any
+  relative or absolute path information originally included in the value. If an agent is
+  using a defaulted (not user-provided) config filename, this default filename value
+  MUST still be provided.
+* This `AgentConfigFile` MUST have a `content_type` of
+  `application/yaml; vendor=splunk; v=1`. This content type tells the remote
+  side how to interpret the content within the `AgentConfigFile` `body`. The v field
+  allows us to revise this format in a backwards compatible way in the future.
+* The `AgentConfigFile` body MUST conform to the body format below:
+
+#### Body Format
+
+The body yaml format MUST match the structure
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -79,7 +79,7 @@ When reporting `EffectiveConfig`, the following MUST be followed:
   `environment_variables`.
 * The `AgentConfigFile` under `environment_variables` MUST have a
   `content_type` of
-  `text/plain; format=properties; vendor=splunk; v=1`. This content type
+  `text/plain; format=properties; vendor=splunk; v=1.0.0`. This content type
   tells the remote side how to interpret the content within the
   `AgentConfigFile` `body`. The v field
   allows us to revise this format in a backwards compatible way in the future.
@@ -145,7 +145,7 @@ environment variables are intentionally omitted from this format.
 #### Example
 
 * name: `environment`
-* content_type: `text/plain; format=properties; vendor=splunk; v=1`
+* content_type: `text/plain; format=properties; vendor=splunk; v=1.0.0`
 
 ```properties
 OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
@@ -178,7 +178,7 @@ When reporting `EffectiveConfig`, the following MUST be followed:
   defaulted (not user-provided) config filename, this default filename
   value MUST still be provided.
 * This `AgentConfigFile` MUST have a `content_type` of
-  `application/yaml; vendor=splunk; v=1`. This content type tells the
+  `application/yaml; vendor=splunk; v=1.0.0`. This content type tells the
   remote side how to interpret the content within the `AgentConfigFile`
   `body`. The v field
   allows us to revise this format in a backwards compatible way in the future.
@@ -238,18 +238,18 @@ tracer_provider:
         exporter:
           otlp_http:
             endpoint: http://localhost:4318/v1/traces
-  meter_provider:
-    readers:
-      - periodic:
-          exporter:
-            otlp_grpc:
-              endpoint: http://localhost:4318/v1/metrics
-  logger_provider:
-    processors:
-      - simple:
-          exporter:
-            otlp_http:
-              endpoint: http://localhost:4318/v1/logs
+meter_provider:
+  readers:
+    - periodic:
+        exporter:
+          otlp_grpc:
+            endpoint: http://localhost:4318/v1/metrics
+logger_provider:
+  processors:
+    - simple:
+        exporter:
+          otlp_http:
+            endpoint: http://localhost:4318/v1/logs
 distribution:
   splunk:
     profiling:
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -48,9 +48,10 @@ agent "effective configuration" in OpAmp payloads. We have made this
 distinction because the two approaches are not fully interoperable. Put
 another way, there is no complete way to smoothly translate environment
 configs into yaml, nor is there a way to represent all declarative confing
-yaml as environment variables.
-
-<todo: link to examples of these env vars that don't yaml>
+yaml as environment variables. For reference, see
+[this comment](https://github.com/open-telemetry/opentelemetry-specification/issues/3967#issuecomment-2813435817)
+in
+[this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/3967).
 
 Additionally, the [OpAmp spec says](https://opentelemetry.io/docs/specs/opamp/#configuration):
 
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -85,7 +85,7 @@ When reporting `EffectiveConfig`, the following SHOULD be followed:
   allows us to revise this format in a backwards compatible way in the future.
 * The `AgentConfigFile` body MUST conform to the body format below:
 
-Agents MAY choose to map environment configuration into the declarative 
+Agents MAY choose to map environment configuration into the declarative
 format described below.
 
 #### Body Format
@@ -136,7 +136,7 @@ Additional configuration items SHOULD NOT be provided by agents.
 If any of these configuration items is not specified via environment
 variables, then a default value MUST be provided. The value provided
 MUST be semantically equivalent to the currently used value, regardless
-of what was actually supplied in the environment. String values that 
+of what was actually supplied in the environment. String values that
 do not have a default SHOULD provide `null`.
 
 For example, even if the environment contains `SPLUNK_PROFILER_ENABLED=true`,
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -47,7 +47,7 @@ Below are two separate and distinct wire format specifications for representing
 agent "effective configuration" in OpAmp payloads. We have made this
 distinction because the two approaches are not fully interoperable. Put
 another way, there is no complete way to smoothly translate environment
-configs into yaml, nor is there a way to represent all declarative confing
+configs into yaml, nor is there a way to represent all declarative config
 yaml as environment variables. For reference, see
 [this comment](https://github.com/open-telemetry/opentelemetry-specification/issues/3967#issuecomment-2813435817)
 in
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -138,9 +138,7 @@ but the agent could not enable the profiler due to runtime platform limitations,
 then the value reported in effective config MUST be
 `SPLUNK_PROFILER_ENABLED=false`.
 
-If an agent has been configured to use more than one endpoint for any
-of its signals, the agent MUST only report the first one. Multivalued
-environment variables are intentionally omitted from this format.
+Multivalued environment variables are intentionally omitted from this format.
 
 #### Example
 
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -23,3 +23,128 @@ The `AgentDescription` message includes a set of
 [non-identifying attributes](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionnon_identifying_attributes).
 GDI language agents SHOULD follow the upstream guidance for which
 OpenTelemetry `Resource` attributes belong in this field.
+
+## Effective Configuration
+
+### Agent configuration context
+
+The [OpAMP specification](https://opentelemetry.io/docs/specs/opamp/#effectiveconfig-message)
+includes a mechanism for agents to report their effective configuration to the server.
+We interpret "effective configuration" to mean the current, active configuration being used by
+the agent, regardless of how that configuration was obtained, injected, defaulted, or
+otherwise evaluated.
+
+Because configuration can come in many forms, this creates several challenges.
+Configuration can be provided as environment variables,
+[declarative config](https://opentelemetry.io/docs/languages/sdk-configuration/declarative-configuration/) files (yaml),
+java system properties, and eventually remote configuration. That's at least 4 ways of
+controlling agent behavior, but some agents will have even more.
+
+Below are two separate and distinct wire format specifications for representing
+agent "effective configuration" in OpAmp payloads. We have made this distinction
+because the two approaches are not fully interoperable. Put another way, there is
+no complete way to smoothly translate environment configs into yaml, nor is there a
+way to represent all declarative confing yaml as environment variables.
+
+<todo: link to examples of these env vars that don't yaml>
+
+Additionally, the [OpAmp spec says](https://opentelemetry.io/docs/specs/opamp/#configuration):
+
+> The configuration of the Agent is a collection of named configuration files
+
+This is somewhat confusing because it suggests the existence of an actual physical 
+file that exists on disk, which isn't always the case. We interpret this use of the word "file" 
+here to refer to a block of content, and not to a physical file stored on disk. 
+
+In the spec below, for simplicity, we treat java system properties as environment
+variables (they are indeed
+[interchangeable](https://opentelemetry.io/docs/zero-code/java/agent/configuration/#configuring-with-environment-variables)).
+
+## Effective Environment Config
+
+When an agent has been started _without_ declarative configuration, it is assumed that its
+configuration is derived from environment variables and default values. This section only applies
+to agents that were started _without_ declarative configuration.
+
+When reporting `EffectiveConfig`, the following MUST be followed:
+
+* The `ConfigMap` MUST contain an `AgentConfigFile` under the name `environment_variables`.
+* The `AgentConfigFile` under `environment_variables` MUST have a `content_type` of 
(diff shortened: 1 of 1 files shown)


specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+
+Some of these effective configuration values have a defined location in
+the yaml structure, while some are Splunk specific.
+
+#### Required fields
+
+The following uses "dot path" shorthand for the equivalent yaml fields.
+The following configuration items SHOULD be reported in the body:
+
+* `tracer_provider.exporter.otlp_http.endpoint`
```
pmcollins, line 209: Do we want `.processors[<n>].<processor>` here and under logger provider as well?

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+#### Example
+
+```yaml
+tracer_provider:
+  processors:
+    - batch:
+        exporter:
+          otlp_http:
+            endpoint: http://localhost:4318/v1/traces
+  meter_provider:
```
pmcollins: nit: the meter_provider and logger_provider blocks should shift left two spaces

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+then the value reported in effective config MUST be
+`SPLUNK_PROFILER_ENABLED=false`.
+
+If an agent has been configured to use more than one endpoint for any
+of its signals, the agent MUST only report the first one. Multivalued
+environment variables are intentionally omitted from this format.
+
+#### Example
+
+* name: `environment`
```
pmcollins, line 151: Do we want this to be `environment_variables`?

pmcollins (approved):
LGTM. Thanks for doing this -- appreciate the simplicity at this stage. Added a few minor comments.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+Because configuration can come in many forms, this creates several challenges.
+Configuration can be provided as environment variables,
+[declarative config](https://opentelemetry.io/docs/languages/sdk-configuration/declarative-configuration/)
+files (yaml), java system properties, and eventually remote
+configuration. That's at least 4 ways of controlling agent behavior, but
+some agents will have even more.
+
+Below are two separate and distinct wire format specifications for representing
+agent "effective configuration" in OpAmp payloads. We have made this
+distinction because the two approaches are not fully interoperable. Put
```
aunshc, line 48: Having gone through the otel spec PR below, it is called out that there is no clean mapping of the set of env vars into declarative config because of fundamental modeling differences. 

Is there an option to define keys for these upstream or via our distro so that complete translation is possible? Or was this an intentional design decision?

Looking at this from a point of view to find a path to a single format. This would avoid the server having to detect + render in two different layouts.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+another way, there is no complete way to smoothly translate environment
+configs into yaml, nor is there a way to represent all declarative confing
```
aunshc:
```suggestion
configs into yaml, nor is there a way to represent all declarative config
```

specification/opamp_datamodel.md
aunshc, line 209: Good starting point. My thought is that we should plan to expand this set. Considering the value effective configuration brings in context of detecting any misconfigurations, reflecting what is being used at that point - users could very likely look to verify the value of another configuration option outside these.

Could we send other items part of the json schema (eg. resource, propagator) i.e. send the full yaml?

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+The following configuration items MUST be reported in the body:
+
+* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
+* `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
+* `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`
+* `SPLUNK_PROFILER_ENABLED`
+* `SPLUNK_PROFILER_MEMORY_ENABLED`
+* `SPLUNK_SNAPSHOT_PROFILER_ENABLED`
+* `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL`
+* `SPLUNK_PROFILER_CALL_STACK_INTERVAL`
```
robsunday, line 130: Do you see a value in reporting additionally `SPLUNK_PROFILER_LOGS_ENDPOINT`?

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+variables, then a default value MUST be provided. The value provided
+MUST be semantically equivalent to the currently used value, regardless
+of what was actually supplied in the environment.
+
+For example, even if the environment contains `SPLUNK_PROFILER_ENABLED=true`,
+but the agent could not enable the profiler due to runtime platform limitations,
+then the value reported in effective config MUST be
+`SPLUNK_PROFILER_ENABLED=false`.
+
+If an agent has been configured to use more than one endpoint for any
```
robsunday: I think this paragraph is not needed when we report differently Environment and Declarative Configs. There is no way to specify multiple endpoints for specific signal with Env Config.

ysolomchenko (commented):
The way I see it, the effective config should not show what the user configured, but rather what actually exists in the application (more like a status view). Since this is the case, we are moving away from the model of “show exactly what the user configured” toward a model that reflects the actual state of the application.

Because there are multiple ways to configure it, as @breedx-splk mentioned, I think we need a single submission model instead of two separate ones. I would suggest using something like a YAML-based model, since it is much more universal than a key-value model.

We could reuse the existing YAML format from the file-based config and just extend it with the missing keys, even if they are represented as key-value pairs at the root level of the file.

At the same time, the front-end should not display the raw YAML, but instead show more user-friendly blocks

(even the same ones that are already planned for the remote config).

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+When reporting `EffectiveConfig`, the following MUST be followed:
+
+* The `ConfigMap` MUST contain an `AgentConfigFile` under the name that
+  exactly matches the filename provided in the `OTEL_CONFIG_FILE`
+  environment variable. This MUST include any relative or absolute path
+  information originally included in the value. If an agent is using a
+  defaulted (not user-provided) config filename, this default filename
+  value MUST still be provided.
+* This `AgentConfigFile` MUST have a `content_type` of
+  `application/yaml; vendor=splunk; v=1`. This content type tells the
```
robsunday: I'd use semantic versioning for more flexibility in the future: `v=1.0.0`

breedx-splk: sure, that's fine....but I still believe that semantic versioning is a lie. 🙃

breedx-splk: You're also deluding yourself if you think that this will ever change more than a few times. Like, ever.

specification/opamp_datamodel.md
breedx-splk: I chose `environment` because I prefer the brevity. If everyone else loves `environment_variables` better, than I am happy to change...but I don't think the extra suffix buys us anything really.

specification/opamp_datamodel.md
breedx-splk: oh good catch! Not even a nit, but a legit classic yaml formatting problem.

specification/opamp_datamodel.md
robsunday: Shouldn't we update it if otel schema version is updated?

specification/opamp_datamodel.md
pmcollins: I mentioned it because line 79 specifies `environment_variables` -- or is that specifying something else?

specification/opamp_datamodel.md
breedx-splk: Not yet. We can expand into that if there's a need. 👍🏻

specification/opamp_datamodel.md
breedx-splk: oh, thanks for pointing out that inconsistency! I think I prefer `environment` so I'll change it above.

specification/opamp_datamodel.md
breedx-splk: I'm not sure what you mean because you quoted from two different paragraphs here. Which one are you referring to?

specification/opamp_datamodel.md
breedx-splk: It was not asked for by @aunshc so I did not include it. I'll defer to him to answer this question. For now, I'd err on the side of keeping things small and building on the foundations. We can always add the profiler endpoint later, if we think it's useful.

specification/opamp_datamodel.md
breedx-splk: I think that rather than trying to send everything from the start, we should start with this minimal set and expand it incrementally.

Yes, we could conceivably/technically send the entire yaml file from disk when the agent is started with declarative configuration. That is something we could build. It was my understanding, however, that we wanted to be able to report what the agent is actually using. Remember that these are two different things! It's important to remember this.

The spec I've put forth here tries to express this -- the agent is reporting its "effective config" for the things we care about. I think it would be a mistake to try and report everything. I'm not even sure there is a comprehensive list of everything anyway.

breedx-splk:
> The way I see it, the effective config should not show what the user configured, but rather what actually exists in the application (more like a status view).

Yes, we are in agreement here. This is why I expressly included the section about what the interpretation of "effective config" means.

>  Since this is the case, we are moving away from the model of “show exactly what the user configured” toward a model that reflects the actual state of the application.

State of the agent, but yes. This was the intention behind the original design of "effective config" and @tigrannajaryan has confirmed this.

> Because there are multiple ways to configure it, as @breedx-splk mentioned, I think we need a single submission model instead of two separate ones. I would suggest using something like a YAML-based model, since it is much more universal than a key-value model.

Why do you think this? Why would you suggest that because there are multiple ways of doing something that there should be a single way of reporting the results? This feels forced. It feels like shoehorning two very different looking things into a uniform box.
 
> We could reuse the existing YAML format from the file-based config and just extend it with the missing keys, even if they are represented as key-value pairs at the root level of the file.

There are many things that we _could_ do. You are free to propose a more "universal" spec and if people prefer that then we'll go with that instead. I'm not married to this approach, but I'm skeptical about a "one-size-fits-all" solution.
 
> At the same time, the front-end should not display the raw YAML, but instead show more user-friendly blocks

I'm not trying to address anything about the backend, storage, or UI with this spec. I'm trying to get alignment on how agents report effective config. How the UI displays something is another issue entirely.

specification/opamp_datamodel.md
breedx-splk: > Is there an option to define keys for these upstream or via our distro so that complete translation is possible? Or was this an intentional design decision?

It's intentional because the long-term direction that OTel is heading is to reduce the use of env vars in favor of declarative config (yaml). This has been confirmed by the otel tech committee many times now. ([ref1](https://github.com/open-telemetry/opentelemetry-specification/pull/3948#issuecomment-2025825640)) ([ref2](https://github.com/open-telemetry/opentelemetry-specification/issues/3967))
 
> Looking at this from a point of view to find a path to a single format. This would avoid the server having to detect + render in two different layouts.

No matter what format is proposed, someone will complain about it and ask the obvious question of "why can't it just match the declarative config schema?" 🙄

specification/opamp_datamodel.md
aunshc: I agree with expanding from here on but I do want move the framing towards what the user could come across and want - 

1. Hard to predict which field could matter for a user's journey. A fixed list of fields works until someone needs another field to diagnose an issue.
2. Users can't see what they didn't set (including defaults) for options outside this set. If that doesn't show up in effective config, they can't tell what their agent is actually using. Would be easier to track this from the UI than having to dig through documentation.

Thinking aloud -> another way of reporting what the agent is actually using is to report the full resolved state - configurations it is running with and has values for, including defaults (similar to the Collector's effective configuration implementation). That then leaves it to the UI/UX to handle effectively (group, collapse, show key ones upfront as needed). 

If we curate a moving list in the spec, every time a field becomes relevant and is useful, it's a spec change, agent implementations and a release. That would be a lag between user needing this vs getting to see it via the API/UI.

For the comprehensive list, could we think of 1) The OTel JSON schema + 2) unmappable env vars ([reference](https://github.com/open-telemetry/opentelemetry-specification/issues/3967#issuecomment-2813435817))  + 3) Splunk specific configurations. This would allow us to keep this bounded and known. Would that suffice? Do let me know if I'm missing a point here.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+The following uses "dot path" shorthand for the equivalent yaml fields.
+The following configuration items SHOULD be reported in the body:
+
+* `tracer_provider.exporter.otlp_http.endpoint`
+* `meter_provider.readers[<n>].periodic.exporter.otlp_http.endpoint`
+* `logger_provider.exporter.otlp_http.endpoint`
+* `distribution.splunk.profiling.always_on.cpu_profiler.sampling_interval`
+* `distribution.splunk.profiling.always_on.memory_profiler`
+* `distribution.splunk.profiling.callgraphs.sampling_interval`
+
```
aunshc, line 217: If we do expand on the settings looking forward, it would be good to include a section on how secrets will be handled i.e. should be redacted. That is a customer expectation based on feedback we have received. Totally understand that this could be handled at agent or collector level but it is required.

ysolomchenko:
> > Because there are multiple ways to configure it, as @breedx-splk mentioned, I think we need a single submission model instead of two separate ones. I would suggest using something like a YAML-based model, since it is much more universal than a key-value model.
> 
> Why do you think this? Why would you suggest that because there are multiple ways of doing something that there should be a single way of reporting the results? This feels forced. It feels like shoehorning two very different looking things into a uniform box.

As you mentioned earlier, the effective config represents the resulting status/state, so it probably shouldn’t depend on how the configuration was originally defined.

Because of that, separating this into two reporting models feels like unnecessary complexity to me, while the end result conceptually is still the same effective configuration.

That’s why I was leaning toward a single, more universal submission model rather than two separate ones.

The current `key-value` approach works well while the configuration surface is still relatively small and flat. My concern is more about how this evolves once we introduce more complex scenarios, for example multiple exporters, no-code itc.

From my perspective, it’s better to design with that future in mind now, while the abstraction boundary is still simple, rather than introducing complexity.

> >We could reuse the existing YAML format from the file-based config and just extend it with the missing keys, even if they are represented as key-value pairs at the root level of the file.
> 
>There are many things that we could do. You are free to propose a more "universal" spec and if people prefer that then we'll go with that instead. I'm not married to this approach, but I'm skeptical about a "one-size-fits-all" solution.

I think this is actually a case where having multiple solutions would be worse than having a single universal one.

We are not really dealing with fundamentally different domains here (“one-size-fits-all” would make more sense in cases where the underlying concepts are unrelated). In this case, different configuration methods still converge into the same effective configuration

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+The following configuration items MUST be reported in the body:
+
+* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
+* `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
+* `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`
+* `SPLUNK_PROFILER_ENABLED`
+* `SPLUNK_PROFILER_MEMORY_ENABLED`
+* `SPLUNK_SNAPSHOT_PROFILER_ENABLED`
+* `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL`
+* `SPLUNK_PROFILER_CALL_STACK_INTERVAL`
```
lachmatt, line 130: Should we include `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL` and `SPLUNK_PROFILER_CALL_STACK_INTERVAL` conditionally, only if respective profiling types are enabled (same goes for other similar settings we might want to include in the future)? I see you suggest something different in your example - it reports `0` for `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL` in such case (if we decide to go that path, should we define values to use to mark corresponding settings as having no effect?)

specification/opamp_datamodel.md
robsunday: I'm sorry for confusion. I ment lines 141-143. These lines should be removed.

seemk:
I think there should only be 1 (wire)format for the configuration and we've let the double format run too far.

I like the stripped down yaml, where agents actually report what they have configured, doesn't matter where it comes from. The only problem is that we can't attach the configuration source to it (env, programmatic, config file) - although this is not really necessary yet and could later be sent as a separate mapping. E.g. for the Node.js distro we could configure all the OTel and Splunk parts while keeping a registry that allows tracking where a given configuration value comes from.

If there exist a set of env vars are not translatable, we are just pushing the problem upstream to our backend/frontend. Would users look at different user interfaces depending on the config format used?

So my suggestion is we only **keep the sparse yaml as a wire format that is populated based on actual configuration and let's get rid of the env var mapping completely.**

specification/opamp_datamodel.md
robsunday: I think iterative approach is a must here. There are hundreds of options to configure. Some are trivial, but some are not, and dedicated, sometimes possibly complex logic will have to be implemented to report the real values.
If we forget to implement a dedicated logic for some property then report will contain false information.
From user perspective - when investigating config issue it is better not to have some information in the report, than have a false information.

With iterative approach we can focus on reliable reporting of the most important items in the beginning, and then add more (maybe even before GA)

breedx-splk:
Ok, after kinda amassing feedback about this initial PR, I'm hearing a few themes:

* Multiple asks for a single format. Nobody supporting separate formats. 
* Putting secrets on the wire is bad and dangerous. 🙀 
* Iterative development is too restrictive, we need everything, everywhere, all at once. 😉 

So to address these, I've reworked this into #395. That is _not_ the "minimal" yaml that @seemk has asked for, because providing the minimal config is in conflict with the other ask to provide everything without having to release. Sorry, we can't have it both ways...that is, unless we add another configuration item for it.... 🤔 j/k.

So anyway, please review #395 and comment there and we'll figure out what we like.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+Note: `<value-char>` is any character that's not a newline.
+
+The order of these lines is not important. Each `<key>` MUST only
+appear in the body once.
+
+#### Required fields
+
+The following configuration items MUST be reported in the body:
+
+* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
```
iskiselev, line 123: what if otlp exporter disabled? are we reporting it with otlp (and will not report in that case) or with other protocol - and it may be possible in that case.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+
+If any of these configuration items is not specified via environment
+variables, then a default value MUST be provided. The value provided
+MUST be semantically equivalent to the currently used value, regardless
+of what was actually supplied in the environment.
+
+For example, even if the environment contains `SPLUNK_PROFILER_ENABLED=true`,
+but the agent could not enable the profiler due to runtime platform limitations,
+then the value reported in effective config MUST be
+`SPLUNK_PROFILER_ENABLED=false`.
```
iskiselev, line 145: It is hard - no we make it dependent not on configuration system, but on state of each component. What if we enable top layer but something gone wrong later and we disabled it inside due to error?

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+in
+[this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/3967).
+
+Additionally, the [OpAmp spec says](https://opentelemetry.io/docs/specs/opamp/#configuration):
+
+> The configuration of the Agent is a collection of named configuration files
+
+This is somewhat confusing because it suggests the existence of an
+actual physical file that exists on disk, which isn't always the case.
+We interpret this use of the word "file" here to refer to a block of
```
tigrannajaryan, line 62: Yes, that's the intended meaning. Please feel free to suggest wording improvements to OpAMP spec.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+and default values. This section only applies to agents that were
+started *without* declarative configuration.
+
+When reporting `EffectiveConfig`, the following MUST be followed:
+
+* The `ConfigMap` MUST contain an `AgentConfigFile` under the name
+  `environment`.
+* The `AgentConfigFile` under `environment` MUST have a
+  `content_type` of
+  `text/plain; format=properties; vendor=splunk; v=1.0.0`. This content type
```
tigrannajaryan, line 82: Do we need the `vendor` pair? Does the format need to be splunk-specific?

Also, do you know if there are any conventions for naming parameters? Is there a common parameter for version to use instead of `v`?

tigrannajaryan: Do you know if there is a well-known MIME type we can use for key/value pairs instead of `text/plain` so that we don't have to define the body format ourselves?

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,246 @@ The `AgentDescription` message includes a set of
...
+configuration.
+
+When running with declarative configuration, the "effective
+configuration" body will follow a similar structure, but will be a
+subset of what the actual file supports.
+
+When reporting `EffectiveConfig`, the following MUST be followed:
+
+* The `ConfigMap` MUST contain an `AgentConfigFile` under the name that
+  exactly matches the filename provided in the `OTEL_CONFIG_FILE`
```
tigrannajaryan, line 181: What happens if the file name is `environment`? How do we resolve the conflict?

tigrannajaryan (commented):
I am not entirely sure about the idea that we separately report config from env vars and config from declarative config file.

How is the user supposed to know what the result of merging those is? 

One important purpose of effective config reporting is for the user to understand what really is the final set of parameters the agent runs with. By reporting the 2 configs separately we are punting the problem down the road for the user to figure out. Now they need to examine the behavior of the agent to understand how these 2 configs are merged and override each other.

specification/opamp_datamodel.md
breedx-splk: > Do we need the vendor pair? Does the format need to be splunk-specific?

I mean, until this stuff is standardized in upstream, I think we need the vendor, yes. Even then, I think "properties" is too generic anyway.

> Do you know if there is a well-known MIME type we can use for key/value pairs instead of `text/plain` so that we don't have to define the body format ourselves?

I'm not aware of one, which is why we landed on this here.

breedx-splk:
> I am not entirely sure about the idea that we separately report config from env vars and config from declarative config file.
> 
> How is the user supposed to know what the result of merging those is?
> 
> One important purpose of effective config reporting is for the user to understand what really is the final set of parameters the agent runs with. By reporting the 2 configs separately we are punting the problem down the road for the user to figure out. Now they need to examine the behavior of the agent to understand how these 2 configs are merged and override each other.

I've clearly failed to make this obvious, or there's something else getting in the way of a shared understanding here...but I'll try and clarify:

There is no merging.

When agents run with declarative config, they ONLY run with declarative config, and environment vars are no longer used to configure the sdk/agent. In this configuration, ONLY yaml should be reported. I think that's covered in this spec.

When agents do NOT run with declarative config, they use environment vars to configure the sdk/agent. In this configuration, no yaml is reported, only the properties style env vars.

@tigrannajaryan does that clear this up?

specification/opamp_datamodel.md
breedx-splk: If the profiler is disabled due to error, then the effective configuration must reflect that. Specifically, if the configuration inserted into the agent has it turned on, but it can't be turned on or is later turned off, then the effective configuration should say `SPLUNK_PROFILER_ENABLED=false`.

specification/opamp_datamodel.md
breedx-splk: If you are not exporting with OTLP then you can report `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=none`.

This was one of the config items explicitly asked for by PM, so if it's not set then they probably don't care...but it should be nice to see that it isn't being used. I don't care.

specification/opamp_datamodel.md
breedx-splk: @aunshc hopefully I have made my point with #395 -- that expecting the agents to simply report everything all the time is completely unreasonable and we will not be doing that. I agree with @robsunday here that iterating is the only pragmatic path forward.

specification/opamp_datamodel.md
breedx-splk: No response, so I'm assuming it's not important for now. Let's keep it out and add it later as needed.

specification/opamp_datamodel.md
breedx-splk: Oh, I see what you mean now. How about I just keep that last sentence "Multivalued
environment variables are intentionally omitted from this format." because that was what I think was actually important....

specification/opamp_datamodel.md
breedx-splk: I don't care about that edge case. We can tell users not to use a file called `environment` if we need to, but we also have the mime type to hint how to handle the thing.

specification/opamp_datamodel.md
breedx-splk: We can do that in a follow-up PR. Please open a tracking issue so that we can address this in the future. Right now, there should not typically be any secrets in the set of required config items, so I think it's reasonably safe to omit this for now and not overcomplicate this already entirely overcomplicated tornado of opinion.

specification/opamp_datamodel.md
breedx-splk: No, we do not want to make these conditional, because it adds complexity. Let's report them always. If profiling is not enabled and you need to report the interval, use the default or make something up.

specification/opamp_datamodel.md
aunshc: Thank you @breedx-splk. In the case that this is not reasonable and reliable to report as explained to me earlier, we should focus on the following configuration use cases for v1:
1. Exporter endpoint settings
2. Profiling
3. Call graphs
4. Sampling
5. Propagation
6. OpAMP 

With a subset of configurations for the top 2 use cases covered in this PR, it would be useful to add the others. Since the semantics/naming are different across env vars and the yaml format - not adding each setting here but let me know if that is better to be shared. 

As called out [above](https://github.com/signalfx/gdi-specification/pull/391#discussion_r3256697564), I still believe we should evolve towards one consistent wire format instead of two. We do not have to get held further by this but it would be important to consider. I'm not tied to the format (yaml or key-value pairs as file), reading through the feedback around using the former - there will be expectation set via docs and UI/UX that effective config is not assumed as declarative config that can be reapplied directly. The source metadata will be useful to report in this single format but can be addressed in a future version as needed.

cc @ysolomchenko @seemk

specification/opamp_datamodel.md
iskiselev: I still beleive we mixing two things here:
- effective configuration - configuration in config system after default values was applied and multi-level config (if ever be supported) was merged
- operational status report / runtime configuration / active configuration - state for each component of our system - based on configuration, interpritation of this configuration by component and some environmental settings.

For example, in AppD DOTNET profiler we have 3 options for inlining switch:
- auto - let us decide if it should be enabled (always when safe) or disabled (on runtimes with known bugs)
- on - always on, even when usafe
- off - always off, even if it can be safely used

In my opinion, effective configuration may have all 3 values: `auto`, `on`, `off`. While status report may have only two values: `on` or `off ` - as `auto` should be resolved into either.

We can make `effective configuration` easier than `operational status report`. And, IMHO, in many cases customer is interested in first - not the second.
Despite neither of this set can be applied as working configuration without carefull review of each setting, `effective configuration` may be more sutable if customer may want to reuse parts of it for configuring other apps.

My opinion - we should not do  operational status report - only efffective config - so, if due to any environmental issue profiler was disabled, but config enabled it - we should still report it as enabled.

If we will need operational status report - it is worse to design it as another feature.

tigrannajaryan:
> There is no merging.
> 
> When agents run with declarative config, they ONLY run with declarative config, and environment vars are no longer used to configure the sdk/agent. In this configuration, ONLY yaml should be reported. I think that's covered in this spec.
> 
> When agents do NOT run with declarative config, they use environment vars to configure the sdk/agent. In this configuration, no yaml is reported, only the properties style env vars.
> 
> @tigrannajaryan does that clear this up?

@breedx-splk yes, that helps and is important. I didn't realize that. I think the proposal is fine if that's the case. I don't mind if different config sources are reported differently, provided that only one is used at a time and the server doesn't need to merge multiple sources or do anything special to figure out the real effective config.

ysolomchenko:
> > I am not entirely sure about the idea that we separately report config from env vars and config from declarative config file.
> > How is the user supposed to know what the result of merging those is?
> > One important purpose of effective config reporting is for the user to understand what really is the final set of parameters the agent runs with. By reporting the 2 configs separately we are punting the problem down the road for the user to figure out. Now they need to examine the behavior of the agent to understand how these 2 configs are merged and override each other.
> 
> I've clearly failed to make this obvious, or there's something else getting in the way of a shared understanding here...but I'll try and clarify:
> 
> There is no merging.
> 
> When agents run with declarative config, they ONLY run with declarative config, and environment vars are no longer used to configure the sdk/agent. In this configuration, ONLY yaml should be reported. I think that's covered in this spec.
> 
> When agents do NOT run with declarative config, they use environment vars to configure the sdk/agent. In this configuration, no yaml is reported, only the properties style env vars.
> 
> @tigrannajaryan does that clear this up?

For .NET this is not true. Some configuration still must come from env vars even with YAML config enabled, due to .NET runtime / profiler startup constraints and because some settings are required before the application starts:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/file-based-configuration.md#not-supported-configurations

breedx-splk:
> For .NET this is not true. Some configuration still must come from env vars even with YAML config enabled, due to .NET runtime / profiler startup constraints and because some settings are required before the application starts:
> 
> https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/file-based-configuration.md#not-supported-configurations

Thanks for the clarification @ysolomchenko. I know this will be a pain point, once again demonstrating that one size often doesn't fit all.

We will need to come up with some provisions/workarounds/exceptions for this.

iskiselev:
I think the central design question is whether this feature reports:

1. **Effective configuration**: final configuration values after defaults, precedence, and config-source merging are applied.
2. **Operational status / active runtime state**: what the agent/components are actually doing after interpreting configuration and accounting for runtime/environmental conditions.

That choice affects most of the remaining design.

In a complex configuration system, effective configuration is the merged result of multiple configuration layers: built-in defaults, app config, machine config, environment overrides, remote overrides, and so on. It is not necessarily identical to any original source file, but it still lives inside the configuration system. Component logic for handling runtime errors, unavailable platform capabilities, environmental constraints, or adaptive behavior is outside that boundary.

Even though we do not currently have a highly layered configuration system, the distinction still exists. With YAML config, for example, we already allow environment variable substitution inside the file. That means the baked effective config can differ from the source YAML even before we consider future config layers.

The tricky part is defaults. Today, some defaults may live inside components and only be applied when a config value is missing. Those can be handled in a few ways:

- Initially hardcode default handling in the effective-config gatherer for the closed list of reported settings.
- Move default value processing into the configuration system.
- Expose an API that lets components register default-value setters with the configuration system.

Any default value not handled by one of those options is component behavior, not effective configuration. Reporting that value would require the operational-status route or a separate feature.

From the end-user perspective, effective configuration and operational status serve different workflows. If users mainly want a status/debug view, then operational status is more useful. If they want to understand the effective configured intent, then effective configuration is closer, even though it still cannot be used as a 1:1 configuration template or remote-config payload. Inferred/defaulted values and environment-derived values already mean the original source shape may be lost. But operational status goes further: it may also hide intentionally configured values that were meant to be resolved by runtime/tooling logic.

For example, if a setting supports `auto`, `on`, and `off`, effective config can report `auto`. Operational status would report only the resolved runtime result, such as `on` or `off`. Both are useful, but they answer different questions.

This also connects to @aunshc's earlier feedback. I read his concern as wanting the reported config to be useful beyond a narrow status/debug view: users should be able to inspect misconfiguration, verify important settings, and potentially reuse parts of the reported configuration with care. At the same time, he also called out that effective config should not be documented or presented as a declarative config payload that can be reapplied directly.

The discussion in #395 seems to confirm that the opposite extreme does not work either. Asking agents to report the entire active configuration or all potentially unknown component state is not realistic, especially with open/plugin-style configuration surfaces. So I do not think the choice is between "tiny fixed list forever" and "report everything now." A more practical v1 is a closed initial list with clear semantics, plus an extensibility path for adding reportable settings later.

One reason I lean toward effective configuration is implementation cost. It seems like the fastest path for the initial version, and also the most natural path if we later want to expose this as an open/extensible API. The configuration system already has the right ownership boundary: it can expose selected effective values without requiring every runtime component to implement a separate status-reporting path.

Making this open/extensible may be harder if the spec imposes strict format requirements, especially YAML-shape requirements. In that model, a component would not only need to register a reportable value, but also describe how that value maps into multiple configuration formats. That adds work and makes extension points more coupled to presentation/wire-format choices.

My preference:

- Use **effective configuration** unless PM has a strong need for operational status.
- Allow both wire formats, and let the agent choose either as an implementation detail.
- Start with a closed set of reported settings, but design it with extensibility in mind.
- For each reported setting from the initial closed list, require the effective-config logic to know the same default/fallback logic used by the component when the value is not explicitly defined.
- Avoid requiring each component to describe the same value in multiple wire/config formats.
- Do not require the effective-config collector to query the component for actual runtime status. If we need that, it should be a separate operational-status feature.

specification/opamp_datamodel.md
breedx-splk: Thanks for the input @iskiselev . I recognize that it's a challenge to resolve the may different flavors of config -- what's coming into the agent, how it's fully resolved internally, and how we report the effective config, all separate things, I agree.

> if due to any environmental issue profiler was disabled, but config enabled it - we should still report it as enabled.

I understand where you're coming from and the complexities that go into this, but what you're suggesting is NOT what product has asked for. I also think that it contradicts the wording in this specification, which @tigrannajaryan has confirmed is aligned with the intention of the upstream OpAMP specification. Effective config is what we put on the wire, and should be a reflection of what the agent is currently operating with (which you have dubbed an "operational status report"). 

I agree that what we put on the wire (effective config) is different than what the agent is actually running with (no good name for this, but operational config or "active config" is probably fine). It is different in that "effective config" is a marshalled format that fits within OpAMP, while each agent has its own internal model(s) for it's active/running/current/operational configuration.

If we were to only report "incoming" (injected) config in the effective config, this would be a mistake. It would hide defaults, overrides, programatic corrections (like the example you provided), and other aspects of reality.

breedx-splk:
> effective configuration is the merged result of multiple configuration layers: built-in defaults, app config, machine config, environment overrides, remote overrides, and so on.

@iskiselev with this statement, you are directly contradicting and changing the semantics of what I have already written in this specification. The fact that we have to even discuss this shows a gap in the upstream OpAMP definition of "effective config", which is why I wrote this in this very PR:

>We interpret "effective configuration" to mean the current, active
configuration being used by the agent, regardless of how that
configuration was obtained, injected, defaulted, or otherwise evaluated.

My intent with this is to report what the agent is currently using.

If you would like to suggest adding words to this to clarify that it includes fallbacks, escape hatches, evaluation of environmental constraints, yadda yadda yadda, we can do that, but I worry that at some point too many words will only serve to cloud an already confusing/complex topic.

I also worry that teasing apart "effective configuration" into this kinda subtle distinction is leading us down a path where we will be then asked to provide both. And I won't want to provide both. I want to start by providing the minimally useful thing so that we can start iterating.

tigrannajaryan:
Can you guys check what the Collector is doing when it is reporting the effective config? It has most of the same concerns (multiple source config files to merge, defaults). If it is doing something reasonable we can mimic that behavior for language SDKs. The desire to be consistent with the Collector can help make this decision.

breedx-splk:
> Can you guys check what the Collector is doing when it is reporting the effective config? It has most of the same concerns (multiple source config files to merge, defaults). If it is doing something reasonable we can mimic that behavior for language SDKs. 

I'm open to precise, constructive wording or format changes in this PR until I merge this on Friday. 

I'm disenchanted by suggestions that we should change directions completely, write an entirely new format, reinterpret what the OpAMP spec says, report everything, etc...so maybe we don't go down that road this late in the game.

> The desire to be consistent with the Collector can help make this decision.

What was the question again?

tigrannajaryan:
> What was the question again?

@breedx-splk I am saying look at the prior art (Collector).

breedx-splk:
> > What was the question again?
> 
> @breedx-splk I am saying look at the prior art (Collector).

That's a weird question.

iskiselev:
I checked the Collector prior art because it seems like the closest OpAMP reference point for this discussion and @tigrannajaryan suggested it. This Collector prior-art check was prepared with AI assistance. I validated the references myself, but may be missing Collector-specific details.

<details>
<summary>
The Collector appears to report **effective configuration** from the configuration layer, not from component runtime state.
</summary>

More concretely, Collector seems to use the same typed config object for component construction and for effective-config reporting.

The config unmarshaler creates each component config by calling `factory.CreateDefaultConfig()` and then unmarshalling user config on top of that default object:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/otelcol/internal/configunmarshaler/configs.go#L21-L51

For example, the batch processor defines non-zero defaults like `send_batch_size`, `timeout`, and `metadata_cardinality_limit` in its factory:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/processor/batchprocessor/factory.go

The OTLP exporter does the same for timeout, retry, queue, compression, and gRPC client defaults:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/exporter/otlpexporter/factory.go

Then `collector.go` gets that typed config, marshals it back into `confmap.Conf`, and passes both into `service.New`: `CollectorConf: conf` for effective config reporting, and `cfg.Receivers` / `cfg.Processors` / `cfg.Exporters` / etc. for component creation:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/otelcol/collector.go#L186-L224

`service.New` stores `CollectorConf` and also builds component builders from the typed config maps:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/service/service.go#L106-L123

Those builders pass the typed config into component factories, e.g. receivers:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/service/internal/builders/receiver.go#L30-L46

Finally, `Service.Start` calls `NotifyConfig(ctx, srv.collectorConf)` on extensions before starting pipelines:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/service/service.go#L240-L255

Extensions that implement `ConfigWatcher` receive that effective config:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/extension/extensioncapabilities/interfaces.go#L39-L45

The OpAMP extension then serializes that `confmap.Conf` as `EffectiveConfig`:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c7a646f913a46b45d4e182cc7ab5b768973258c3/extension/opampextension/opamp_agent.go

So this is not just raw input config. It is the post-unmarshal typed config with component factory defaults applied, marshaled back into `confmap.Conf`, and reported through the `ConfigWatcher` / OpAMP extension path.
</details>

At the same time, Collector reports runtime state separately. The OpAMP extension reports `ComponentHealth`, and Collector also has Health Check / Health Check V2 extensions for liveness/readiness/runtime status:

- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c7a646f913a46b45d4e182cc7ab5b768973258c3/extension/healthcheckextension/README.md
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c7a646f913a46b45d4e182cc7ab5b768973258c3/extension/healthcheckv2extension/README.md

One useful example is `endpoint: localhost:0`. In Go, port `0` asks the OS to choose an available port. Collector creates the listener from that configured endpoint:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/config/confignet/confignet.go

For OTLP receiver, Collector logs the concrete runtime listener address after bind:
https://github.com/open-telemetry/opentelemetry-collector/blob/009300517f680bae6ebf0640f3d8213ec888e4ab/receiver/otlpreceiver/otlp.go

So if config says `endpoint: localhost:0`, runtime logs may show the selected port, for example `127.0.0.1:53742`. But I do not see that concrete runtime port being written back into effective config. Effective config remains the configured intent, `localhost:0`; the selected port is runtime state.

That is the distinction I am trying to preserve here.

I agree this slightly changes the interpretation of "effective config" currently proposed in this PR, but I do not think it violates OpAMP. OpAMP has separate fields/capabilities for `effective_config` and `health`, so it does not require effective config to also be a full runtime status report.

The practical change to this PR can still be minimal. I do not think we need to redesign the feature or introduce a new payload shape. I think this mostly comes down to changing a few words so that the spec does not require the effective-config reporter to track actual component runtime state.

In many real cases the difference between a config-layer effective value and an observed runtime value may not matter at all. But when it does matter, the stronger requirement is harder to implement correctly and easier to get subtly wrong. An agent may either delay compliance because it has to query component internals, or report something that looks like runtime truth but is only partially resolved.

My suggested direction is therefore narrower:

1. Effective config should report the config-layer effective value for the selected/reported settings, including source precedence, env interpolation, and known defaults/fallbacks where the agent config system can compute them.
2. It should not require the reporter to query component runtime state or convert configured intent into observed operational outcome.
3. The two wire formats should both remain acceptable, and choosing one should be an implementation detail.
4. If we need "what is actually running right now?" later, we can add a separate runtime status feature, or explicitly extend this feature later with stronger requirements.

The goal of this change is not to lower the usefulness of the feature. It is to make the first compliant implementation faster, less error-prone, and easier to extend later. If product needs runtime status as a separate user-facing capability, we should keep that door open, but I am not sure we should force that harder requirement into effective config v1.

seemk:
I'm ok with the yaml part of the PR, but I can't understand that when declarative config is not used why do we need an additional env var / key-value based wire format? Can't we just generate the yaml based on the config received from env? 

Based on the example env config in the PR, the same example YAML can be used.

It's not an issue for the Node.js distro, just seems redundant.

ysolomchenko:
> * Allow both wire formats, and let the agent choose either as an implementation detail.

Let the agent decide which format will be used. I agree with that. (But don't say that if the agent uses env vars, we should report key-value format, and if the agent uses YAML, we should report YAML.)

breedx-splk:
> I'm ok with the yaml part of the PR, but I can't understand that when declarative config is not used why do we need an additional env var / key-value based wire format? Can't we just generate the yaml based on the config received from env?

I thought I had addressed this question already in the PR body itself, but let me quote it here for clarity:

>  We have made this distinction because the two approaches are not fully interoperable. Put another way, there is no complete way to smoothly translate environment configs into yaml, nor is there a way to represent all declarative config yaml as environment variables. For reference, see [this comment](https://github.com/open-telemetry/opentelemetry-specification/issues/3967#issuecomment-2813435817) in [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/3967).

I think you're right, though -- the current set of configs doesn't have a conflict, so I think I'm ok with relaxing the MUST into a SHOULD, and then the agents can decide. 

I don't see a huge problem with that, but it does surface the fact that the content type is kinda serving double duty -- it describes the format but also tells the receiver how the agent was initialized. Having that as implicit/implied is stupid, so I think we should also include the `OTEL_CONFIG_FILE` and `OTEL_EXPERIMENTAL_CONFIG_FILE` (yes, deprecated) in the required set of configs. 

What that then means is that we will have to explicitly spec where this goes in the yaml (diverging from upstream declarative config), and how to represent a null/missing value...because if your agent is configured w/ environment vars and you are sending yaml, you'll need to indicate that a declarative config file was not configured. I don't _love_ that but it's fine.

I'll make these edits, thanks for the feedback @seemk @ysolomchenko .

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,256 @@ The `AgentDescription` message includes a set of
...
+actively used endpoints.
+
+Implementations MAY choose to include disabled/inactive features in the yaml
+document with commented out lines.
+
+#### Example
+
+```yaml
+otel.config.file: /usr/otel/agent.yaml
+otel.experimental.config.file: null
```
ysolomchenko:
```suggestion
otel_config_file: /usr/otel/agent.yaml
otel_experimental_config_file: null
```

ysolomchenko (approved):

specification/opamp_datamodel.md
```diff
@@ -125,27 +128,28 @@ The following configuration items MUST be reported in the body:
 * `SPLUNK_SNAPSHOT_PROFILER_ENABLED`
 * `SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL`
 * `SPLUNK_PROFILER_CALL_STACK_INTERVAL`
+* `OTEL_CONFIG_FILE`
```
robsunday, line 131: These env vars may contain absolute path that in a limited way exposes filesystem structure of the host to the server. It can be considered as a sensitive data.
Should we care now?

robsunday (approved):
Looks good for now. Just minor comments.

specification/opamp_datamodel.md
breedx-splk: I do not care right now. I agree we should give this some future consideration.

specification/opamp_datamodel.md
breedx-splk: Thanks @ysolomchenko I think that's a good catch.

PR #375, opened 2026-02-23
OpAMP configuration for agents

breedx-splk:
This allows us to align configuration for agents that support OpAMP.

* opt-in / enable
* url to connect to
* (optional) polling interval

PR #393, opened 2026-05-19
Adjust poll_interval to the more realistic value.

breedx-splk:
Following up on [this comment](https://github.com/signalfx/gdi-specification/pull/375#discussion_r3263720628) from #375.

I generally prefer examples that aren't identical to the defaults, because I figure that people will just copy/paste them...but sure, whatever.


Kielek (approved):

Kielek:
> I generally prefer examples that aren't identical to the defaults, because I figure that people will just copy/paste them...but sure, whatever.

Yeah, copy-pasting is the clue here. It is better to have some working/almost defaykt value, than some strage 12, which could lead to very frequent requests,

robsunday (approved):

pellared (approved):

PR #395, opened 2026-05-20
Effective config spec [redux]

breedx-splk:
Based on feedback in #391, let's make another attempt at representing agent effective config.

Specifically, compared to #391, this:

* only provides one format
* introduces some vague idea of masking secrets
* tells agents to provide their entire active configuration, so that we don't ever have to iterate. 🙄

### DIFF
--- a/specification/effective_config_example.yaml
+++ b/specification/effective_config_example.yaml
@@ -2,6 +2,29 @@
 file_format: "1.0"
 disabled: false
 log_level: "info"
+distribution:
+  splunk:
+    access_token: "***********"
+    realm: "us0"
+    trace_response_header:
+      enabled: true
+    profiling:
+      exporter:
+        otlp_log_http:
+          endpoint: "http://127.0.0.1:4318/v1/logs"
+      always_on:
+        cpu_profiler:
+          enabled: true
+          sampling_interval: 1000
+        memory_profiler:
+          enabled: false
+      callgraphs:
+        enabled: false
+        sampling_interval: 10
+        selection_probability: 0.01
+    "opamp:/development":
+      endpoint: "https://opamp.example.com/v1/opamp"
+      polling_interval: 30000
 attribute_limits:
   attribute_count_limit: 128
   attribute_value_length_limit: 12000
@@ -24,7 +47,7 @@ tracer_provider:
             compression: "gzip"
             endpoint: "http://127.0.0.1:4318/v1/traces"
             headers_list:
-              - "x-sf-token=*"
+              - "x-sf-token=***********"
             timeout: 10000
             tls:
               ca_file: ""
@@ -46,7 +69,7 @@ meter_provider:
             default_histogram_aggregation: "explicit_bucket_histogram"
             endpoint: "http://127.0.0.1:4318/v1/metrics"
             headers_list:
-              - "x-sf-token=*"
+              - "x-sf-token=*************"
             temporality_preference: "cumulative"
             timeout: 10000
             tls:
@@ -65,7 +88,7 @@ logger_provider:
             compression: "gzip"
             endpoint: "http://127.0.0.1:4318/v1/logs"
             headers_list:
-              - "x-sf-token=*"
+              - "x-sf-token=**********"
             timeout: 10000
             tls:
               ca_file: ""
@@ -75,53 +98,18 @@ logger_provider:
         max_export_batch_size: 512
         max_queue_size: 2048
         schedule_delay: 1000
-distribution:
-  splunk:
-    access_token: "*************"
-    realm: "us0"
-    trace_response_header:
-      enabled: true
-    profiling:
-      exporter:
-        otlp_log_http:
-          endpoint: "http://127.0.0.1:4318/v1/logs"
-      always_on:
-        cpu_profiler:
-          enabled: true
-          sampling_interval: 1000
-        memory_profiler:
-          enabled: false
-      callgraphs:
-        enabled: false
-        sampling_interval: 10
-        selection_probability: 0.01
-    "opamp:/development":
-      endpoint: "https://opamp.example.com/v1/opamp"
-      polling_interval: 30000
 "instrumentation/development":
   general:
     stability_opt_in_list: []
   java:
     activej_http:
       enabled: true
-    activej_http_6:
-      "0":
-        enabled: true
     akka_actor:
       enabled: true
-    akka_actor_2:
-      "3":
-        enabled: true
     akka_actor_fork_join:
       enabled: true
-    akka_actor_fork_join_2:
-      "5":
-        enabled: true
     akka_http:
       enabled: true
-    akka_http_10:
-      "0":
-        enabled: true
     akka_http_client:
       enabled: true
     akka_http_server:
@@ -130,104 +118,48 @@ distribution:
       enabled: true
     alibaba_druid:
       enabled: true
-    alibaba_druid_1:
-      "0":
-        enabled: true
     annotations:
       enabled: true
     apache_dbcp:
       enabled: true
-    apache_dbcp_2:
-      "0":
-        enabled: true
     apache_dubbo:
       enabled: true
-    apache_dubbo_2:
-      "7":
-        enabled: true
     apache_elasticjob:
       enabled: true
       experimental_span_attributes: false
-    apache_elasticjob_3:
-      "0":
-        enabled: true
     apache_httpasyncclient:
       enabled: true
-    apache_httpasyncclient_4:
-      "1":
-        enabled: true
     apache_httpclient:
       enabled: true
-    apache_httpclient_2:
-      "0":
-        enabled: true
-    apache_httpclient_4:
-      "0":
-        enabled: true
-    apache_httpclient_5:
-      "0":
-        enabled: true
     apache_shenyu:
       enabled: true
       experimental_span_attributes: false
-    apache_shenyu_2:
-      "4":
-        enabled: true
     armeria:
       enabled: true
-    armeria_1:
-      "14":
-        enabled: true
-      "3":
-        enabled: true
     armeria_grpc:
       enabled: true
-    armeria_grpc_1:
-      "14":
-        enabled: true
     async_http_client:
       enabled: true
-    async_http_client_1:
-      "8":
-        enabled: true
-      "9":
-        enabled: true
-    async_http_client_2:
-      "0":
-        enabled: true
     avaje_jex:
       enabled: true
-    avaje_jex_3:
-      "0":
-        enabled: true
     aws_lambda:
       enabled: true
       flush_timeout: 10000
     aws_lambda_core:
       enabled: true
-    aws_lambda_core_1:
-      "0":
-        enabled: true
     aws_lambda_events:
       enabled: true
-    aws_lambda_events_2:
-      "2":
-        enabled: true
     aws_sdk:
       enabled: true
       experimental_record_individual_http_error: false
       experimental_span_attributes: false
       experimental_use_propagator_for_messaging: false
     aws_sdk_1:
-      "11":
-        enabled: true
       "11_core":
         enabled: true
       "11_sqs":
         enabled: true
     aws_sdk_2:
-      "2":
-        enabled: true
       "2_bedrock_runtime":
         enabled: true
       "2_core":
@@ -240,55 +172,23 @@ distribution:
         enabled: true
     axis2:
       enabled: true
-    axis2_1:
-      "6":
-        enabled: true
     azure_core:
       enabled: true
-    azure_core_1:
-      "14":
-        enabled: true
-      "19":
-        enabled: true
-      "36":
-        enabled: true
-      "53":
-        enabled: true
     c3p0:
       enabled: true
-    c3p0_0:
-      "9":
-        enabled: true
     camel:
       enabled: true
       experimental_span_attributes: false
-    camel_2:
-      "20":
-        enabled: true
     cassandra:
       enabled: true
-    cassandra_3:
-      "0":
-        enabled: true
-    cassandra_4:
-      "0":
-        enabled: true
-      "4":
-        enabled: true
     clickhouse:
       enabled: true
     clickhouse_client:
       enabled: true
     clickhouse_client_v1:
       enabled: true
-    clickhouse_client_v1_0:
-      "5":
-        enabled: true
     clickhouse_client_v2:
       enabled: true
-    clickhouse_client_v2_0:
-      "8":
-        enabled: true
     common:
       db_statement_sanitizer:
         enabled: true
@@ -319,67 +219,22 @@ distribution:
     couchbase:
       enabled: true
       experimental_span_attributes: false
-    couchbase_2:
-      "0":
-        enabled: true
-      "6":
-        enabled: true
-    couchbase_3:
-      "1":
-        "6":
-          enabled: true
-        enabled: true
-      "2":
-        enabled: true
-      "4":
-        enabled: true
     cxf:
       enabled: true
-    cxf_3:
-      "0":
-        enabled: true
-      "2":
-        enabled: true
     dropwizard_metrics:
       enabled: false
-    dropwizard_metrics_4:
-      "0":
-        enabled: true
     dropwizard_views:
       enabled: true
-    dropwizard_views_0:
-      "7":
-        enabled: true
     elasticsearch:
       capture_search_query: false
       enabled: true
       experimental_span_attributes: false
     elasticsearch_api_client:
       enabled: true
-    elasticsearch_api_client_7:
-      "16":
-        enabled: true
     elasticsearch_rest:
       enabled: true
-    elasticsearch_rest_5:
-      "0":
-        enabled: true
-    elasticsearch_rest_6:
-      "4":
-        enabled: true
-    elasticsearch_rest_7:
-      "0":
-        enabled: true
     elasticsearch_transport:
       enabled: true
-    elasticsearch_transport_5:
-      "0":
-        enabled: true
-      "3":
-        enabled: true
-    elasticsearch_transport_6:
-      "0":
-        enabled: true
     executors:
       enabled: true
       include: []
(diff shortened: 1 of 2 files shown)


specification/opamp_datamodel.md
```diff
@@ -23,3 +23,129 @@ The `AgentDescription` message includes a set of
...
+
+### Agent configuration context
+
+The
+[OpAMP specification](https://opentelemetry.io/docs/specs/opamp/#effectiveconfig-message)
+includes a mechanism for agents to report their effective configuration
+to the server.
+We interpret "effective configuration" to mean the current, active
+configuration being used by the agent, regardless of how that
+configuration was obtained, injected, defaulted, or otherwise evaluated.
```
iskiselev, line 37: So, spec talk only about configuration, not about how it will be interpreted by component?
So, I read it - if configuration processor set sime default value on missing setting - it should be reported.
If component use some default for key not presented in config - it is behaviour and not configuration, it should not be reported.

It is very important difference.
We can report all configuration as extandable entity. Ig some plugin put new key in it - it will be reported (here we have a problem - currently at least in .NET there is no explicit semantic for security sensitive config. I don't know how we can filter-out sensitive setting which we don't know about - unless we extend config API later, which will be not soon.

Alternatively, we can report how some features/component work. That is closed set of reportable items. We can define which is minimum set, but each agent will decide what it want to report.

I beleive that, at least for .NET best format is key=value - it is internal config representation. Nothing in spec said that we should be able to automatically merge back effective config.

Additionally, we may want to report how config file looked (source and processed) - in that case it will be raw format used by files. It should not be interpreted on backed. It is just for refence.
We can include both, as separate file:
- effective (key value pair)
- src/<file_path> (yaml without substitutes env, key/value pair with existing env variables otherwise)
- processed/<file_path> (yaml with resolved env references, key/value pair with existing env variables otherwise)

That still will be compatible with what I read from OpAmp spec. And all of it is effective config.

specification/effective_config_example.yaml
ysolomchenko, line 1: This example shows what will be in the future. For now, I’d focus on what we need for the first version.

https://github.com/breedx-splk/gdi-specification/blob/eb84be2b7303f93293563e9571d0b02be5a70760/specification/opamp_datamodel.md?plain=1#L234-#L262

specification/opamp_datamodel.md
ysolomchenko, line 37:
```suggestion
We interpret "effective configuration" to mean the actual configuration currently 
used by the application at runtime, regardless of how that configuration was obtained.
```

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,129 @@ The `AgentDescription` message includes a set of
...
+files (yaml), java system properties, and eventually remote
+configuration. That's at least 4 ways of controlling agent behavior, but
+some agents will have even more.
+
+We've decided to collapse these mechanisms into a single
+"effective configuration" wire format in OpAmp payloads. This decision
+was made in spite of the fact that some configuration will be difficult
+(or nearly impossible) to represent in this format.
+It is believed that having one universal grab bag format will be better
+than two domain-specific formats.
```
ysolomchenko, line 51: This information is unnecessary because it does not matter how or from where the configuration was obtained; what matters is the configuration that is actually being used by the application.

```suggestion

```

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,129 @@ The `AgentDescription` message includes a set of
...
+
+#### Filtering Sensitive Data
+
+The body of the effective configuration MUST NOT include secrets or other
+sensitive data. If a configuration being used by an agent does truly include
+sensitive values, then the agent MUST overwrite these values with the asterisk
+character (`*`) before the effective config is sent over OpAMP.
+
+The vague and nebulous description of what constitutes "sensitive data"
+or a "secret" is outside the scope of this specification.
```
ysolomchenko, line 133: At the beginning, if we leave only the `Required fields`, then Sensitive Data is not included in them.
```suggestion

```

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,129 @@ The `AgentDescription` message includes a set of
...
+This yaml MUST provide a blend of splunk-specific configuration and all
+of the existing upstream configurations that exist. This includes all
+effective config values for:
+
+* every possible sdk configuration item
+* every configuration item for every possible instrumentation
+* every single default value for every possible configuration item that has
+  not been specified by other configuration (environment, declarative,
+  programmatic, etc.)
+* every possible feature flag and experimental feature
```
ysolomchenko, line 104: I think that might happen in the future, but for now we should focus only on the most essential ones.

specification/opamp_datamodel.md
ysolomchenko: I like the idea of sending what we have as well, but in a slightly different format:

- effective config (the YAML file containing the actual configuration currently used by the application)
- raw env vars (values used for configuration)
- raw YAML (YAML without substituted environment variables)

robsunday:
Before we switch to single format approach I’d like to bring up one more issue.
If agent is configured with env vars then it will be rather impossible to switch to declarative config (used by remote config) on the fly without updating startup scripts/link/dockercompose/etc.
We can safely assume users will have to make some additional effort to switch to declarative config before using remote config.
Before it happens, users may need to have ability to investigate config issues, right? In such a case a key=value format of effective config is much better, as it matches 1:1 with their agent config. What would be the benefit of reporting effective config in yaml format for env vars config?

If we would like to convert complete env vars config into yaml declarative config format, then it will be really significant effort, especially if we would like to convert everything in a non iterative way. In my opinion it is not worth the effort, unless we are going to create a migration tool for the customers.

For declarative config the effective config should be reported in yaml format, as recommended in this PR.

RassK:
> Before we switch to single format approach I’d like to bring up one more issue. If agent is configured with env vars then it will be rather impossible to switch to declarative config (used by remote config) on the fly without updating startup scripts/link/dockercompose/etc. We can safely assume users will have to make some additional effort to switch to declarative config before using remote config. 

Absolutely true 👍 

> Before it happens, users may need to have ability to investigate config issues, right? In such a case a key=value format of effective config is much better, as it matches 1:1 with their agent config. What would be the benefit of reporting effective config in yaml format for env vars config?

I'm fairly sure that if those configurations look similar, then somebody assumes those are copy-edit-pastable. In reality those are not[^1], otherwise what's the point of a separate spec.

Currently seems the wire format and display (UI) format is mixed up. Wire format should be whatever is common to use for such cases (easy to construct, easy to parse). UI format should be whatever is best from the UX perspective.

> If we would like to convert complete env vars config into yaml declarative config format, then it will be really significant effort, especially if we would like to convert everything in a non iterative way. In my opinion it is not worth the effort, unless we are going to create a migration tool for the customers.

Agree, this is more of a standalone (web)tool.

> tells agents to provide their entire active configuration, so that we don't ever have to iterate. 🙄

In the end if we send full yaml or env config, then there is no point of effective configuration anyway? There should be a clear goal of the effective config, something like:
- replace variables for better general overview (does not need a custom spec)
- eliminate config conflicts / overlapping sections (that changes the structure, as keys are evaluated, reduced, added, etc)
- expand important defaults (not all are important I guess)
- should report the source to answer the important why question? (does change the structure, as is additional metadata)
- protects secrets (does need a schema to understand what are secrets)

So IMO that should revert some process how we look into this spec (CC: @aunshc):
- Find out which questions the effective config should answer (defines the data)
- Find out what do you want to display out (defines the metadata)
- Find out the spec that unifies the data and metadata experience (how to gather, how to send, what to send, ...)

[^1]: Yaml may look secure for now of conflicting sections but that doesn't mean future will be the same. Env config is definitely not safe and some keys conflict with others.

specification/effective_config_example.yaml
```diff
@@ -0,0 +1,786 @@
...
+        max_export_batch_size: 512
+        max_queue_size: 2048
+        schedule_delay: 1000
+"instrumentation/development":
+  general:
+    stability_opt_in_list: "http/dup,database,messaging/dup"
+  java:
+    activej_http:
+      enabled: true
+    akka_actor:
```
Kielek, line 172: Just remove this line and everyting below?

BTW `enabled: true` is against configuration SIG recommendation.

ysolomchenko:
> > Before it happens, users may need to have ability to investigate config issues, right? In such a case a key=value format of effective config is much better, as it matches 1:1 with their agent config. What would be the benefit of reporting effective config in yaml format for env vars config?

as it matches 1:1 with their agent config - is not really true. The effective config is not the same as the config users originally provide, so the exact output format is not important.

> I'm fairly sure that if those configurations look similar, then somebody assumes those are copy-edit-pastable. In reality those are not[1](#user-content-fn-1-e80813f1feaaa9a48c978df7e1b52389), otherwise what's the point of a separate spec.
> 
> Currently seems the wire format and display (UI) format is mixed up. Wire format should be whatever is common to use for such cases (easy to construct, easy to parse). UI format should be whatever is best from the UX perspective.
> 

Could we use the prepared UI for remote (apply) config, but for effective config?
https://www.figma.com/design/PtcS5PfRqCplxMx520oZ4Y/OTel-Collector-Management?node-id=5505-62078&t=uZMGGzWSerETS0D1-0

I agree with @RassK that if we only show YAML to users, they may assume it is a declarative config and try to copy-paste it directly. That would blur the line between the wire format and the user-facing representation.

But this is another issue.

specification/effective_config_example.yaml
breedx-splk: Please give a citation.

While that may be true, we need to report this stupid junk...otherwise someone won't even realize that there is a setting that they can turn on or off. 🤷🏻 I hate it too.

RassK: > Please give a citation.
> 
> While that may be true, we need to report this stupid junk...otherwise someone won't even realize that there is a setting that they can turn on or off. 🤷🏻 I hate it too.

That should be true for configuration. But effective configuration is not configuration but more like a markup.

But now the real question is if wire format should exclude it and display format include it (based on schema).

specification/opamp_datamodel.md
breedx-splk: Let's address some of @iskiselev's questions/concerns:

> So, spec talk only about configuration, not about how it will be interpreted by component?

I don't know...but regardless, based on the number of times folks have confused "the yaml file" with "what the agent is actually using" I wanted the spec here to absolutely flipping clear. These are two different things. When we say effective config here, we intend it to be what the agent is actively using. Not anything else.

> So, I read it - if configuration processor set sime default value on missing setting - it should be reported. 

Yes. This spec says that all config, even defaults, should be reported.

> If component use some default for key not presented in config - it is behaviour and not configuration, it should not be reported.

No. This spec says it should be reported. Report it. If you don't send it, then users may never know that the setting exists.

> It is very important difference. We can report all configuration as extandable entity. Ig some plugin put new key in it - it will be reported (here we have a problem - currently at least in .NET there is no explicit semantic for security sensitive config. I don't know how we can filter-out sensitive setting which we don't know about - unless we extend config API later, which will be not soon.

Yup. We don't have a definition for "sensitive", so I just threw my hands up and said that it's needed but not (yet?) defined here. I hate it too.

> Alternatively, we can report how some features/component work. That is closed set of reportable items. We can define which is minimum set, but each agent will decide what it want to report.

In the first PR, people argued against this. There was pushback that incrementally determining what users might one day care about would be too time consuming. 🤷🏻 

> I beleive that, at least for .NET best format is key=value - it is internal config representation. Nothing in spec said that we should be able to automatically merge back effective config.

The internal representation _should_ not be considered for this discussion. We're only trying to spec out how the internal representation gets surfaced on the wire. Every language has a different way of modeling their config while it's in memory.

> Additionally, we may want to report how config file looked (source and processed) - in that case it will be raw format used by files. It should not be interpreted on backed. It is just for refence. We can include both, as separate file:

We are not being asked to report the source of configuration. We're only being asked to report the effective configuration. I will treat that as an entirely different problem.

iskiselev: @breedx-splk, if we need to report component behaviour (how it will react on non-configured setting) - our configuration system should know about every component and how they work in different modes or how to check it. It is impossible with open infrastructure, not at least in current state of it - component new version may always add new config / change deafults or some plugin can be installed. It may be possible if some config reporting API will be developed and each component will be able to report settings back to it - but even in that case, it will be not all settings - but all reported settings.
Alternative aproach is reporting setting files without parsing them. File are designed to be extendable and new settings not known to config system may be added there (or, alternatively, special prefixed env variables).

It is possible to have some hybrid aproaches - when config files are augmented to *know set of defaults / runtime overrides*.

But we can't ask report all, potentially unknown state.

breedx-splk: > It is impossible with open infrastructure, not at least in current state of it 
<snip>
> But we can't ask report all, potentially unknown state.

I agree. I think it's completely ridiculous to ask this. I think the best we could do is to hard code a list of "well here is everything we know today" and then revisit that periodically, but that also doesn't sound sustainable.

specification/opamp_datamodel.md
breedx-splk: It is intended to provide some additional context behind the rationale.

specification/opamp_datamodel.md
breedx-splk: I agree.....focusing on the essential ones would be the better approach.

The challenge is that I tried that in the other PR. Some people are concerned that if we don't send the entire set that there will be UI use cases that cannot be explored. I still think we should start with a minimal set and expand as needed, but there's been push back.

specification/opamp_datamodel.md
seemk: >We are not being asked to report the source of configuration. We're only being asked to report the effective configuration. I will treat that as an entirely different problem.

Agree, let's keep it very simple and add features iteratively.

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,129 @@ The `AgentDescription` message includes a set of
...
+  value MUST still be provided.
+* This `AgentConfigFile` MUST have a `content_type` of
+  `application/yaml; vendor=splunk; v=1.0.0`. This content type tells the
+  remote side how to interpret the content within the `AgentConfigFile`
+  `body`. The v field
+  allows us to revise this format in a backwards compatible way in the future.
+* The `AgentConfigFile` body MUST conform to the body format below, and
+  agents MUST NOT send effective configuration in any other format because
+  it weakens our insistence on uniformity and because we are worried that
+  it could cause backend services to do more work.
```
seemk, line 87: If I understand it correctly, we'd key the effective configuration by the file path?

I would normalize the keys and use a standard one for the effective configuration, e.g. `config.effective` -> `AgentConfigFile` and keep the config file path out of this PR.

Then (out of the scope of this PR) if we later actually need to send the env vars or the YAML file (as a debugging aid) we can just add additional keys:
```
config.raw.yaml.path -> myconfig.yaml
config.raw.yaml.content -> ditto
config.raw.env -> ditto
config.effective.origins -> mapping of keys to where the configuration value came from in config.effective (default / env / yaml / programmatic)
```

specification/opamp_datamodel.md
```diff
@@ -23,3 +23,129 @@ The `AgentDescription` message includes a set of
...
+
+Agents MUST provide effective configuration for every possible
+configuration item, including features that have been disabled or
+for configurations that include empty maps, dictionaries, or lists,
+unless the mere existence of a configuration in the yaml changes the
+semantics of what is currently "effectively" used.
+
+When the existence of a configuration item in the yaml changes the
+semantics of what's "effective", implementations SHOULD still include
+these configs as commented out yaml line(s) (beginning with a `#`).
```
seemk, line 118: Do we have an example where / how this could happen?

specification/opamp_datamodel.md
seemk: Yeah, context is nice!

specification/opamp_datamodel.md
seemk: I'd keep the masking part, make it configurable (default: on) in the OpAMP configuration.
Afaik only `X-SF-TOKEN` would make sense to filter from our side.

Out of scope of this PR: Agents could add an option to let users add a programmatic filter/mask for the keys, but only as a suggestion (SHOULD).

seemk:
> Before we switch to single format approach I’d like to bring up one more issue. If agent is configured with env vars then it will be rather impossible to switch to declarative config (used by remote config) on the fly without updating startup scripts/link/dockercompose/etc. We can safely assume users will have to make some additional effort to switch to declarative config before using remote config. Before it happens, users may need to have ability to investigate config issues, right? In such a case a key=value format of effective config is much better, as it matches 1:1 with their agent config. What would be the benefit of reporting effective config in yaml format for env vars config?
> 
> If we would like to convert complete env vars config into yaml declarative config format, then it will be really significant effort, especially if we would like to convert everything in a non iterative way. In my opinion it is not worth the effort, unless we are going to create a migration tool for the customers.
> 
> For declarative config the effective config should be reported in yaml format, as recommended in this PR.

This is about the wire format for us, the remote config can be in whatever format we want that is convenient and will most likely be a small subset of features to toggle at first (profiling, callgraphs). It does not require that the agent is configured with declarative config.

Implementation detail / not really relevant for this PR: we don't need to convert env vars to yaml. Agents should keep a configuration registry that is queried when building the SDK components. The registry knows about the configuration sources and which one to prioritize.

seemk:
> In the end if we send full yaml or env config, then there is no point of effective configuration anyway? 

No, the point of effective configuration is to see the actual applied configuration. You can overwrite the configuration values programmatically and you'd be sending config that does not represent reality.

specification/effective_config_example.yaml
breedx-splk: This was intended to be a response to the complaint that adding features incrementally could be slower than PM wants...so I just forced everything. It's supposed to be stupid because it is.

breedx-splk:
> If agent is configured with env vars then it will be rather impossible to switch to declarative config (used by remote config) on the fly without updating startup scripts/link/dockercompose/etc. 

@robsunday Literally nobody has ever asked for this or suggested this. I'm not sure why you would bring this up. We're not doing that. Please don't complicate this even more than it already is!

I also have no idea why you say "declarative config (used by remote config)". That's confusing. Remote config is a different thing than declarative config. Please let me know if you need me to explain the differences to you. I think it's a mistake to suggest that declarative config will be or is used by remote config.

> If we would like to convert complete env vars config into yaml declarative config format, then it will be really significant effort, especially if we would like to convert everything in a non iterative way. In my opinion it is not worth the effort, unless we are going to create a migration tool for the customers.

I agree. The approach suggested here is stupid because it was intended to be stupid to prove the point that "just send everything" is a flipping fool's errand.

breedx-splk:
I hope that I have, at least in part, demonstrated that it is completely absurd to assume or ask that we "just send everything always and we'll figure it out in the backend". It's not realistic and the agents aren't built for this kind of thing. We will pursue another option.

PR #396, opened 2026-06-01
Start adding spec for remote config

breedx-splk:
Taking some small baby steps around remote configuration. Trying to do this very incrementally. This is experimental.

This PR:

* Adds an opt-in config item for remote configuration.
* Specs the key name that agents should care about from the remote config map.

What it intentionally does not do, that I expect will need consideration in following PRs:

* content type of the file
* format details of the file
* required/optional config items
* behavioral/operational concerns

### DIFF
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -366,10 +366,11 @@ looks like this:
 ```yaml
 distribution:
   splunk:
-    opamp:/development
+    opamp/development:
       endpoint: http://some.opamp-host.com:3420/v1/opamp
       polling_interval: 30000
-      remote_config: true
+      features:
+        remote_config:
 ```
 
 _Note: When the opamp endpoint node is present in the yaml, it implies that
--- a/specification/configuration.md
+++ b/specification/configuration.md
@@ -346,16 +346,17 @@ MUST be opt-in and not enabled by default.
 When using environment-variable based agent configuration, the following
 MUST be provided when opting in to OpAMP:
 
-| Name                    | Default | Description                                              |
-|-------------------------|---------|----------------------------------------------------------|
-| `SPLUNK_OPAMP_ENABLED`  | false   | Set to `true` to opt into connecting to an OpAMP server. |
-| `SPLUNK_OPAMP_ENDPOINT` | (none)  | The URL endpoint of the OpAMP server to use.             |
+| Name                         | Default | Description                                              |
+|------------------------------|---------|----------------------------------------------------------|
+| `SPLUNK_OPAMP_ENABLED`       | false   | Set to `true` to opt into connecting to an OpAMP server. |
+| `SPLUNK_OPAMP_ENDPOINT`      | (none)  | The URL endpoint of the OpAMP server to use.             |
 
 The following optional configuration options MAY be provided:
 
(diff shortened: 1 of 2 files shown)


ysolomchenko (approved):

robsunday (approved):
Looks good.

specification/configuration.md
```diff
@@ -346,16 +346,17 @@ MUST be opt-in and not enabled by default.
 When using environment-variable based agent configuration, the following
 MUST be provided when opting in to OpAMP:
 
-| Name                    | Default | Description                                              |
-|-------------------------|---------|----------------------------------------------------------|
-| `SPLUNK_OPAMP_ENABLED`  | false   | Set to `true` to opt into connecting to an OpAMP server. |
-| `SPLUNK_OPAMP_ENDPOINT` | (none)  | The URL endpoint of the OpAMP server to use.             |
+| Name                         | Default | Description                                              |
```
Kielek: I doubt that any autoformater will do this.

specification/configuration.md
```diff
@@ -368,6 +369,7 @@ distribution:
     opamp:/development
       endpoint: http://some.opamp-host.com:3420/v1/opamp
       polling_interval: 30000
+      remote_config: true
```
Kielek: Do you want to list all other future settings here?
shouldn't it be under:
capabilities/supported features?

specification/configuration.md
breedx-splk: You noticed the fact that I had initially put the remote config in this table, but realized that it belonged in the table below, and after I removed it the whitespace stayed. 🤷🏻 Is this important? No. Will I revert it back? Yes.

specification/configuration.md
breedx-splk: >Do you want to list all other future settings here?

No, I don't know what the future holds.

>shouldn't it be under:
capabilities/supported features?

I'm not sure what you mean? Can you show me?

Kielek: Just add one more node for the functionaltity configuration, now you are mixing communication layer with other stuff

The second part, the true shouldn't be here. If the node is avaialble, it means that it should be enabled.

```yml
     capabiliteis/feature_flags/functional_config:
          remote_config: # true shouldn't be here at all, if the node is avaialbe it means that is enabled
               foo: any future config for remote_config, eg. only accept partial of settings
```

robsunday: I actually suggested a bit different approach in the upstream: https://github.com/open-telemetry/opentelemetry-configuration/issues/544#issuecomment-4225244736 , but idea is the same - group the settings.
Unfortunately there is not much discussions any longer in this issue.

breedx-splk: Yeah, I suppose I'm late to the upstream discussion, but using the existence of a bare node to imply that a feature is enabled feels completely goofy to me. Because that seems to be the convention, though, I'll change this PR to follow. It will look stupid, but that's fine.

breedx-splk: Ok change made, is it better now at least?

Kielek (approved):

t2t2 (approved):

seemk (approved):

PR #398, opened 2026-06-04
Add data format and semantics for remote configuration.

breedx-splk:
Until we can more reliably leverage telemetry profiles, let's spec this as changes from remote config. 

Users want to be able to:

* turn the profiler on/off
* alter the sampling interval for profiling

We expect this format to be expanded as we add capabilities.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -289,8 +289,38 @@ behavior, MUST accept payloads from the
 [`AgentRemoteConfig.AgentConfigMap`](https://opentelemetry.io/docs/specs/opamp/#agentremoteconfig-message)
 with the key `splunk.remote.config`.
 
+Agents SHOULD attempt to parse the payload of `splunk.remote.config` when the
+content-type is `application/yaml`.
+
 Agents MAY accept payloads from other `AgentRemoteConfig.AgentConfigMap` keys
 not defined in this specification.
 
-* The specific format of this configuration is TBD.
-* The specific configuration settings inside this configuration are TBD.
+### Data Format
+
+When agents receive a remote configuration with the key `splunk.remote.config` and
+content type `application/yaml`, they SHOULD expect the following format:
+
+```yaml
+distribution:
+  splunk:
+    profiling:
+      always_on:
+        cpu_profiler:
+          sampling_interval: 1001
+```
+
+* When `cpu_profiler` is present, it indicates that the CPU profiler should be started
+  if it is not currently running.
+* When `cpu_profiler` is omitted, it indicates that the CPU profiler should be stopped
+  if it is currently running.
+* When `sampling_interval` is present, it indicates how often the profiler should sample.
+
+Agents SHOULD detect when remote configuration differs from current effective configuration
+and SHOULD alter its internal state to match remote config. In other words, remote config
+can cause the agent to start or stop the profiler or modify the sampling interval.
+
+When an agent is able to change state based on remote configuration values,
+subsequent effective configuration reports (as requested by the server)
+MUST reflect these changes.
+
+Agents MUST NOT perform any local persistence of remote configuration values.
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -309,6 +309,8 @@ distribution:
           sampling_interval: 1001
 ```
 
+Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
+
 * When `cpu_profiler` is present, it indicates that the CPU profiler should be started
   if it is not currently running.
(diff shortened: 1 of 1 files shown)


specification/opamp_datamodel.md
```diff
@@ -289,8 +289,40 @@ behavior, MUST accept payloads from the
...
+          sampling_interval: 1001
+```
+
+Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
+
+* When `cpu_profiler` is present, it indicates that the CPU profiler should be started
+  if it is not currently running.
+* When `cpu_profiler` is omitted, it indicates that the CPU profiler should be stopped
+  if it is currently running.
+* When `sampling_interval` is present, it indicates how often the profiler should sample.
```
ysolomchenko: Do you expect that we will always receive the full remote config? Have you considered sending only the settings that need to be changed instead of sending the entire configuration every time?

breedx-splk: I do not expect that we will always receive the full remote config. Sending the entire thing every time feels misguided, but ultimately we won't be in control over what we receive.

ysolomchenko: With this comment:

"When `cpu_profiler` is omitted, it indicates that the CPU profiler should be stopped if it is currently running."

This mean that we always need to send the full configuration. For example, if new elements such as `memory_profiler` are added in the future, how would we send an update for only `cpu_profiler` without affecting memory profiling?

If `memory_profiler` is omitted, wouldn't that imply it should be disabled as well? In that case, it seems we'd always need to send the complete configuration.

robsunday (approved):
LGTM

specification/opamp_datamodel.md
breedx-splk: OH I'm sorry, I completely see your point now. Very valid and useful to call this out.

I see this a side effect of the declarative config format not having (and discouraging) the use of `enabled: true` type fields. Because we're trying to basically follow that same format, we are stuck with this then.

Fortunately, I think we will only have a few dynamic items to start with, and as this set expands over time I think it can still be manageable. So with that, I guess I change my answer -- Yes, I think we should expect the server to send the entire remote config, because the absence of some items would otherwise suggest that they should be disabled.

specification/opamp_datamodel.md
```diff
@@ -289,8 +289,42 @@ behavior, MUST accept payloads from the
 [`AgentRemoteConfig.AgentConfigMap`](https://opentelemetry.io/docs/specs/opamp/#agentremoteconfig-message)
 with the key `splunk.remote.config`.
 
+Agents SHOULD attempt to parse the payload of `splunk.remote.config` when the
+content-type is `application/yaml`.
```
Kielek, line 293: We have made a splunk specific content type for reporting configuration, is it intentional to keep it so generic here?

specification/opamp_datamodel.md
Kielek: Can you add this to this PR and check with our backend team if they are fine?

ysolomchenko: In that case, what serves as the base for a remote config file? Because without a base the backend cannot send a full config file.

specification/opamp_datamodel.md
breedx-splk: Yes, because reporting effective config is not the same as receiving remote configuration. We can enhance this with extra flavor/version information later if you'd like.

specification/opamp_datamodel.md
breedx-splk: >Can you add this to this PR and check with our backend team if they are fine?

@Kielek I'm not sure what you're asking me here. Are you suggesting that there be some statement about the full remote config being present or whatever?

Kielek (approved):

specification/opamp_datamodel.md
Kielek: I mean include information that we expect always full configuration.

breedx-splk: I'm hesitant to add to much language around this, and "we always expect the full config" doesn't tell the full story. Trying to explain it all gets complicated and wordy quickly, so I'm not sure. For example, the value for the sampling rate wouldn't have to be provided every time, if the user just wanted to use the default value. And then it would also require explaining (again!) how the absence of a thing sometimes means it should be turned off (which I still despise, but whatever)....

If people really really want it, I can try and add something to help clarify...but I'm trying not to.

Kielek: I think that we should avoid any room for the interpretation. It should be as strict as possible.

Not a blocker for this PR.

t2t2 (approved):

seemk (approved):

PR #402, opened 2026-07-10
[RemoteConfig] Add Support for Memory profiler

ysolomchenko:
This PR adds support for the Memory Profiler in remote configuration.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -307,6 +307,8 @@ distribution:
       always_on:
         cpu_profiler:
           sampling_interval: 1001
+        memory_profiler:
+          max_memory_samples: 201
 ```
 
 Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
@@ -317,11 +319,17 @@ Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
   if it is currently running.
 * When `sampling_interval` is present, it indicates how often the profiler should
   sample.
+* When `memory_profiler` is present, it indicates that the memory profiler
+  should be started if it is not currently running.
+* When `memory_profiler` is omitted, it indicates that the memory profiler
+  should be stopped if it is currently running.
+* When `max_memory_samples` is present, it indicates the maximum number of
+  memory samples collected per minute.
 
 Agents SHOULD detect when remote configuration differs from current effective
 configuration and SHOULD alter its internal state to match remote config. In
 other words, remote config can cause the agent to start or stop the profiler or
-modify the sampling interval.
+modify profiler settings.
 
 When an agent is able to change state based on remote configuration values,
 subsequent effective configuration reports (as requested by the server)
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -312,16 +312,16 @@ distribution:
 
 Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
 
-* When `cpu_profiler` is present, it indicates that the CPU profiler should be started
+* When `cpu_profiler` is present, it indicates that the CPU profiler SHOULD be started
   if it is not currently running.
-* When `cpu_profiler` is omitted, it indicates that the CPU profiler should be stopped
+* When `cpu_profiler` is omitted, it indicates that the CPU profiler SHOULD be stopped
   if it is currently running.
-* When `sampling_interval` is present, it indicates how often the profiler should
+* When `sampling_interval` is present, it indicates how often the profiler SHOULD
   sample.
 * When `memory_profiler` is present, it indicates that the memory profiler
-  should be started if it is not currently running.
+  SHOULD be started if it is not currently running.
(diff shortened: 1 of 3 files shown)


specification/opamp_datamodel.md
```diff
@@ -317,11 +319,17 @@ Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
   if it is currently running.
 * When `sampling_interval` is present, it indicates how often the profiler should
   sample.
+* When `memory_profiler` is present, it indicates that the memory profiler
+  should be started if it is not currently running.
+* When `memory_profiler` is omitted, it indicates that the memory profiler
+  should be stopped if it is currently running.
+* When `max_memory_samples` is present, it indicates the maximum number of
```
robsunday: I have some concerns regarding `max_memory_samples` property:

- It is not reported with effective config at the moment. We should now keep in-sync remote config and effective config.
- Name, description and probably the technical aspects of this configuration option differs between distributions, so I think we need some broader discussion on it.

ysolomchenko: Looks like it's also .NET-specific. I removed it from the general spec. [0e25cc8](https://github.com/signalfx/gdi-specification/pull/402/commits/0e25cc833d1d511cbd64eb7001407d55e73053de)

robsunday (approved):

t2t2 (approved):

specification/opamp_datamodel.md
```diff
@@ -317,11 +318,15 @@ Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
   if it is currently running.
 * When `sampling_interval` is present, it indicates how often the profiler should
   sample.
+* When `memory_profiler` is present, it indicates that the memory profiler
+  should be started if it is not currently running.
+* When `memory_profiler` is omitted, it indicates that the memory profiler
+  should be stopped if it is currently running.
```
pellared, line 332:
```suggestion
* When `memory_profiler` is present, it indicates that the memory profiler
  SHOULD be started if it is not currently running.
* When `memory_profiler` is omitted, it indicates that the memory profiler
  SHOULD be stopped if it is currently running.

```

pellared (commented):
Can you add a changelog entry?

pellared (approved):

specification/opamp_datamodel.md
ysolomchenko: Fixed in [a3a1942](https://github.com/signalfx/gdi-specification/pull/402/commits/a3a1942b4d6677d0ec1b9b2a76d06ecaa229c51c)

specification/opamp_datamodel.md
```diff
@@ -307,21 +307,26 @@ distribution:
...
-* When `cpu_profiler` is omitted, it indicates that the CPU profiler should be stopped
+* When `cpu_profiler` is omitted, it indicates that the CPU profiler SHOULD be stopped
   if it is currently running.
-* When `sampling_interval` is present, it indicates how often the profiler should
+* When `sampling_interval` is present, it indicates how often the profiler SHOULD
   sample.
+* When `memory_profiler` is present, it indicates that the memory profiler
+  SHOULD be started if it is not currently running.
+* When `memory_profiler` is omitted, it indicates that the memory profiler
+  SHOULD be stopped if it is currently running.
```
MrAlias, line 331: These bullets create a runtime lifecycle for `SPLUNK_PROFILER_MEMORY_ENABLED` that conflicts with the stable Instrumentation Libraries contract in `specification/configuration.md`: the initial setting SHOULD last for the application run, and a deactivated profiler MUST not add profiling overhead.

A process that starts with memory profiling disabled and later receives this payload cannot satisfy both documents. Startup-bound profilers must either ignore the update or retain hooks/callbacks while disabled. Please define the remote-config exception and precedence, including how unsupported transitions and the disabled-state overhead guarantee work, or keep memory-profiler enablement startup-only.

MrAlias (commented):
The remote memory-profiler lifecycle conflicts with the stable configuration contract, so this should not merge until its precedence and disabled-state behavior are defined.

specification/opamp_datamodel.md
ysolomchenko: Fixed in 0bb55c4

MrAlias (approved):

specification/opamp_datamodel.md
```diff
@@ -307,21 +314,26 @@ distribution:
...
 
 Agents SHOULD detect when remote configuration differs from current effective
 configuration and SHOULD alter its internal state to match remote config. In
 other words, remote config can cause the agent to start or stop the profiler or
-modify the sampling interval.
+modify profiler settings.
```
breedx-splk, line 336: I think this was originally intended to provide "for instance" style representative examples of things that could be changed, and you've narrowed the scope to only what's defined here...which I think is fine. It's more precise this way, but slightly less forward looking (will need to change when sampling rate changes come in).

breedx-splk (approved):
This looks good.

PR #403, opened 2026-07-15
First pass spec for taking thread dumps

breedx-splk:
We are being requested to make this quick hack to cause thread dumps to be taken and reported via the profiling signal. I have a number of notable concerns with this, but am trying to make the most out of an unfortunately regettable thing.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -347,7 +347,7 @@ configuration option MAY be provided:
 | `SPLUNK_OPAMP_EXPERIMENTAL_REMOTE_CONTROL` | false   | Set to `true` to accept experimental remote-control commands. |
 
 The equivalent Java system property is
-`splunk.opamp.experimental_remote_control`.
+`splunk.opamp.experimental.remote.control`.
 
 When using declarative configuration, `experimental_control` enables the
 feature:
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -432,8 +432,8 @@ ignored.
 
 ### `thread.dump` Command
 
-The `thread.dump` command, currently supported by the Java agent, requests one
-or more thread dumps of the agent process. Its body has the following format:
+The `thread.dump` command, requests one or more thread dumps of the agent
+process. Its body has the following format:
 
 ```text
 thread.dump
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -357,7 +357,8 @@ distribution:
   splunk:
     opamp/development:
       endpoint: http://some.opamp-host.com:3420/v1/opamp
-      experimental_control: true
+      features:
+        experimental_control:
 ```
 
 OpAMP must be enabled in the agent for this feature to work.
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -328,3 +328,220 @@ subsequent effective configuration reports (as requested by the server)
 MUST reflect these changes.
 
 Agents MUST NOT perform any local persistence of remote configuration values.
+
+## Remote Control
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
+
+Some use cases require an OpAMP server to send an instruction, or command, to a
+running agent. The standard `ServerToAgentCommand` message currently supports
+only agent restarts. Agents MAY support additional commands using the
+experimental transports described in this section. This feature is opt-in and
+MUST NOT be enabled by default.
+
+When using environment-variable based agent configuration, the following
+configuration option MAY be provided:
+
+| Name                                       | Default | Description                                                   |
+|--------------------------------------------|---------|---------------------------------------------------------------|
+| `SPLUNK_OPAMP_EXPERIMENTAL_REMOTE_CONTROL` | false   | Set to `true` to accept experimental remote-control commands. |
+
+The equivalent Java system property is
+`splunk.opamp.experimental_remote_control`.
+
+When using declarative configuration, `experimental_control` enables the
+feature:
+
+```yaml
+distribution:
+  splunk:
+    opamp/development:
+      endpoint: http://some.opamp-host.com:3420/v1/opamp
+      experimental_control: true
+```
+
+OpAMP must be enabled in the agent for this feature to work.
+
+This feature is independent of the call graph profiler (also known as the
+snapshot profiler) and does not require `SPLUNK_SNAPSHOT_PROFILER_ENABLED` to
+be set.
+
+### Command Transport
+
+Commands MAY be carried in an OpAMP `CustomMessage` or, for OpAMP clients that
+do not yet expose custom-message support, in an `AgentRemoteConfig` message.
+Both transports carry the same command body: UTF-8 encoded,
+newline-separated text. An agent MUST parse and execute the body identically
+regardless of which transport carries it.
+
+#### Custom Message Transport
+
+Servers SHOULD use an OpAMP
+[`CustomMessage`](https://opentelemetry.io/docs/specs/opamp/#custom-messages)
+when the client supports it. The server MUST set the fields as follows:
+
+| Field        | Value                                      |
+|--------------|--------------------------------------------|
+| `capability` | `com.splunk.opamp.experimental_command/v1` |
+| `type`       | `command`                                  |
+| `data`       | UTF-8 encoded command body.                |
+
+Agents and servers that support this transport SHOULD advertise
+`com.splunk.opamp.experimental_command/v1` in
+`CustomCapabilities.capabilities`. A server SHOULD send commands using
+`CustomMessage` when the agent advertises this capability.
+
+For example, the `data` field for a command that requests three thread dumps is
+the byte representation of the following text:
+
+```text
+thread.dump
+01JZQ3E4YJYX8N2YQF5Z7K4M6P
+3
+1000
+```
+
+#### Remote Configuration Compatibility Transport
+
+**Status**: [Deprecated](../README.md#versioning-and-status-of-the-specification)
+
+This transport remains supported as a compatibility
+mechanism for OpAMP clients, including the current Java client, that do not yet
+expose custom-message support. A server SHOULD NOT use this transport when the
+agent advertises the `com.splunk.opamp.experimental_command/v1` custom
+capability.
+
+The server MAY place a command in the
+[`AgentRemoteConfig.AgentConfigMap`](https://opentelemetry.io/docs/specs/opamp/#agentremoteconfig-message)
+using the reserved config filename `COMMAND_HACKS`. The value MUST be an
+`AgentConfigFile` whose body is UTF-8 encoded. Its `content_type` SHOULD be
+`text/plain; charset=utf-8`.
+
+When `COMMAND_HACKS` is present, the agent MUST interpret its body as a command
+rather than ordinary remote configuration. A server MAY send a command and
+remote configuration together in the same
+`AgentConfigMap`.
+
+#### Command Body
+
+Command bodies consist of UTF-8 encoded, newline-separated, positional fields.
+The first line is the command name. The meaning and type of subsequent lines
+are defined by the command. Leading and trailing whitespace on each line is
+ignored.
+
+### `thread.dump` Command
+
+The `thread.dump` command, currently supported by the Java agent, requests one
+or more thread dumps of the agent process. Its body has the following format:
+
+```text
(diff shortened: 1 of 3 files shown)


specification/opamp_datamodel.md
```diff
@@ -328,3 +328,220 @@ subsequent effective configuration reports (as requested by the server)
...
+
+When using declarative configuration, `experimental_control` enables the
+feature:
+
+```yaml
+distribution:
+  splunk:
+    opamp/development:
+      endpoint: http://some.opamp-host.com:3420/v1/opamp
+      experimental_control: true
```
ysolomchenko:
```suggestion
    opamp/development:
      endpoint: http://some.opamp-host.com:3420/v1/opamp
        features:
          remote_control:
```

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,220 @@ subsequent effective configuration reports (as requested by the server)
...
+```
+
+OpAMP must be enabled in the agent for this feature to work.
+
+This feature is independent of the call graph profiler (also known as the
+snapshot profiler) and does not require `SPLUNK_SNAPSHOT_PROFILER_ENABLED` to
+be set.
```
ysolomchenko: Shouldn't it be placed in the [configuration file](https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md)?

breedx-splk: maybe, I guess yeah, and then links in both directions or something.

specification/opamp_datamodel.md
breedx-splk: Oh yeah, this declarative config stuff hates explicit feature toggles and prefers empty blocks to do the same thing. Ugh. Regardless, I think it very much needs the word `experimental` in there -- so that it sends a clear message to the user.

specification/opamp_datamodel.md
breedx-splk: This opamp data model file already contains both wire format, behavioral stuff, and config stuff. I think having all the "opamp stuff" in one place is nice. Maybe we should just rename the file from `opamp_datamodel.md` to `opamp.md` or something?

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+
+When using environment-variable based agent configuration, the following
+configuration option MAY be provided:
+
+| Name                                       | Default | Description                                                   |
+|--------------------------------------------|---------|---------------------------------------------------------------|
+| `SPLUNK_OPAMP_EXPERIMENTAL_REMOTE_CONTROL` | false   | Set to `true` to accept experimental remote-control commands. |
+
+The equivalent Java system property is
+`splunk.opamp.experimental_remote_control`.
```
MrAlias: Could we make this `splunk.opamp.experimental.remote.control`? The shared Java mapping in `configuration.md` says environment-variable underscores become `.` or `-`; keeping underscores here creates an exception and leaves Java users with an option the common configuration contract does not define.

```suggestion
`splunk.opamp.experimental.remote.control`.
```

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+agent advertises the `com.splunk.opamp.experimental_command/v1` custom
+capability.
+
+The server MAY place a command in the
+[`AgentRemoteConfig.AgentConfigMap`](https://opentelemetry.io/docs/specs/opamp/#agentremoteconfig-message)
+using the reserved config filename `COMMAND_HACKS`. The value MUST be an
+`AgentConfigFile` whose body is UTF-8 encoded. Its `content_type` SHOULD be
+`text/plain; charset=utf-8`.
+
+When `COMMAND_HACKS` is present, the agent MUST interpret its body as a command
```
MrAlias: Could we make handling `COMMAND_HACKS` conditional on the experimental-control option, and say agents advertise the command capability only when that option is enabled? As written, this unconditional MUST applies to an agent with ordinary remote configuration enabled but remote control still at its default of `false`; receiving this filename would execute a command anyway.

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+#### Command Body
+
+Command bodies consist of UTF-8 encoded, newline-separated, positional fields.
+The first line is the command name. The meaning and type of subsequent lines
+are defined by the command. Leading and trailing whitespace on each line is
+ignored.
+
+### `thread.dump` Command
+
+The `thread.dump` command, currently supported by the Java agent, requests one
```
MrAlias: Could we either land the Java handler with this change or describe this as proposed rather than currently supported? The Java remote-config processor currently recognizes `splunk.remote.config` only, so a `COMMAND_HACKS` entry is ignored. Calling this supported makes the compatibility path look deployable before an implementation exists.

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+```text
+thread.dump
+01JZQ3E4YJYX8N2YQF5Z7K4M6P
+3
+1000
+```
+
+The first thread dump is collected immediately. When `count` is greater than
+one, the agent waits `interval_millis` after completing one thread dump before
+starting the next. The agent MUST use the command's `job_id` for every thread
```
MrAlias, line 425: Could we define duplicate-delivery semantics here? [OpAMP can repeat remote-configuration content](https://opentelemetry.io/docs/specs/opamp/#servertoagent-message) after reconnects or status reports, and `job_id` currently appears only in the output labels. A bounded in-memory record of accepted or completed job IDs, together with defined accepted, rejected, and completed outcomes, would keep a resend from collecting the same sequence again.

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+
+The `LogRecord` uses the standard profiling envelope with the following
+attributes:
+
+| Attribute                          | Value                                            |
+|------------------------------------|--------------------------------------------------|
+| `com.splunk.sourcetype`            | `otel.profiling`                                 |
+| `profiling.data.type`              | `cpu`                                            |
+| `profiling.data.format`            | `pprof-gzip-base64`                              |
+| `profiling.instrumentation.source` | `threaddump`                                     |
```
MrAlias, line 446: Could we update `semantic_conventions.md` alongside this? Its pprof section currently defines `profiling.instrumentation.source` for `snapshot` and `continuous`, but not `threaddump`; without adding this source and its pprof semantics, consumers following the central contract have no defined value to accept.

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+<count>
+<interval_millis>
+```
+
+| Line | Name            | Type                           | Default | Description                                 |
+|------|-----------------|--------------------------------|---------|---------------------------------------------|
+| 1    | command         | string                         | none    | MUST be the literal `thread.dump`.          |
+| 2    | job_id          | string                         | none    | Opaque identifier for the dump request.     |
+| 3    | count           | positive 32-bit signed integer | 1       | Number of thread dumps to collect.          |
+| 4    | interval_millis | positive 32-bit signed integer | 1000    | Delay in milliseconds between thread dumps. |
```
MrAlias: Could we put a maximum on `count` and `interval_millis`, and define cancellation or rejection behavior? The current types accept 2,147,483,647 dumps one millisecond apart. A typo or bad server-side value could otherwise keep an agent generating expensive thread dumps and profiling logs for a very long time.

MrAlias (commented):
Thanks for carrying the compatibility bridge through. I think we need a few contract details resolved before we merge it: the present wording can bypass the opt-in and allow duplicate or unexpected command execution. Once those are nailed down, it should give the Java implementation—and future agents—a dependable target.

specification/opamp_datamodel.md
breedx-splk: Sure.

specification/opamp_datamodel.md
breedx-splk: I have an open PR for this https://github.com/signalfx/splunk-otel-java/pull/2951

breedx-splk: but I agree that the "currently supported" thing is potentially misleading and could be improved.

specification/opamp_datamodel.md
ysolomchenko: Could it be `experimental_remote_control:` or `remote_control/development:` ?

specification/opamp_datamodel.md
ysolomchenko: We already have an OpAMP configuration section:

https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#opamp

So I think it would be better to move the configuration-related part there.

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+
+#### Custom Message Transport
+
+Servers SHOULD use an OpAMP
+[`CustomMessage`](https://opentelemetry.io/docs/specs/opamp/#custom-messages)
+when the client supports it. The server MUST set the fields as follows:
+
+| Field        | Value                                      |
+|--------------|--------------------------------------------|
+| `capability` | `com.splunk.opamp.experimental_command/v1` |
```
robsunday, line 364: This is a very general capability. It is fine solution for the urgent release, but did you also consider using `com.splunk.opamp.experimental_thread_dump/v1` instead?
What I'm suggesting is to advertise supported commands in a granular way, where each supported command has its own capability, its own version and data format.
In the future we will probably support more commands and agent would be able to "handshake" with a server in a stricter way. It would also simplify independent developing of new commands and evolving existing.

specification/opamp_datamodel.md
```diff
@@ -328,3 +328,221 @@ subsequent effective configuration reports (as requested by the server)
...
+The first line is the command name. The meaning and type of subsequent lines
+are defined by the command. Leading and trailing whitespace on each line is
+ignored.
+
+### `thread.dump` Command
+
+The `thread.dump` command, requests one or more thread dumps of the agent
+process. Its body has the following format:
+
+```text
```
robsunday, line 395: What about yaml format here? We use it in many places already.

specification/opamp_datamodel.md
breedx-splk: ok fine.

specification/opamp_datamodel.md
breedx-splk: Although the thing you're describing definitely improves robustness, I would prefer to not have that level of completeness in the spec at this very early stage. I think that the impact of a potential/possible resend is manageable.

specification/opamp_datamodel.md
breedx-splk: just trying to keep things simple and not over-engineer a feature I hope goes away anyway. 🤷🏻

specification/opamp_datamodel.md
breedx-splk: This sounds like an entirely reasonable and pragmatic thing to suggest. It also sounds more complicated and possibly over-engineered for what we're doing right now.

My preference would be for the agents to not receive any commands whatsoever, and for all of this to not even exist. With that, I'm not inclined to try and design a comprehensive, future-proof, extensible solution right now. Just trying to do the basic thing.

I don't think that having additional finer-grained capabilities defined later, even if they too were "commands" or whatever. In fact, if. this stupid thing sticks around and we have to support it long term, that might be a road toward "stability" -- so we keep the "experimental" umbrella for all commands not stable, and the path to stabilizing a given command is to have it as a first-class versioned capability. I dunno. I didn't set out to do that, but maybe that's also reasonable?

specification/opamp_datamodel.md
robsunday: Well, once something is released it takes years to get rid of it...
And I'm afraid that changing format will be equally hard.

breedx-splk: Nah, we can leverage content-type if/when that's needed.

specification/opamp_datamodel.md
robsunday: Ok. Let's see what future brings us

robsunday (approved):
Looks good for now, with one small question

breedx-splk:
Sounds like we're backing down from this strategy for now, in favor of some profiling enhancements.

PR #406, opened 2026-07-24
[RemoteConfig] Add Support for Callgraphs

ysolomchenko:
This PR adds support for the Callgraphs in remote configuration.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -326,8 +326,8 @@ Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
   started if it is not currently running.
 * When `cpu_profiler` is omitted, it indicates that the CPU profiler SHOULD be
   stopped if it is currently running.
-* When `cpu_profiler.sampling_interval` is present, it indicates how often the
-  CPU profiler SHOULD sample.
+* When `cpu_profiler.sampling_interval` is present, it specifies the CPU
+  profiler's sampling interval in milliseconds.
 * When `cpu_profiler.sampling_interval` is omitted but `cpu_profiler` is
   present, it indicates that the CPU profiler is enabled with the default
   sampling interval.
@@ -339,8 +339,8 @@ Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
   be started if it is not currently running.
 * When `callgraphs` is omitted, it indicates that the snapshot profiler SHOULD
   be stopped if it is currently running.
-* When `callgraphs.sampling_interval` is present, it indicates how often the
-  snapshot profiler SHOULD sample call stacks.
+* When `callgraphs.sampling_interval` is present, it specifies the snapshot
+  profiler's call stack sampling interval in milliseconds.
 * When `callgraphs.sampling_interval` is omitted but `callgraphs` is present,
   it indicates that the snapshot profiler is enabled with the default sampling
   interval.
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -315,20 +315,41 @@ distribution:
         cpu_profiler:
           sampling_interval: 1001
         memory_profiler:
+      callgraphs:
+        sampling_interval: 10
+        selection_probability: 0.01
 ```
 
 Agents SHOULD be relaxed in parsing and SHOULD ignore all other values.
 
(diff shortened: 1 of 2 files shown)


seemk (approved):

specification/opamp_datamodel.md
```diff
@@ -315,20 +315,41 @@ distribution:
...
-  sample.
+* When `cpu_profiler` is present, it indicates that the CPU profiler SHOULD be
+  started if it is not currently running.
+* When `cpu_profiler` is omitted, it indicates that the CPU profiler SHOULD be
+  stopped if it is currently running.
+* When `cpu_profiler.sampling_interval` is present, it indicates how often the
+  CPU profiler SHOULD sample.
+* When `cpu_profiler.sampling_interval` is omitted but `cpu_profiler` is
+  present, it indicates that the CPU profiler is enabled with the default
+  sampling interval.
```
robsunday, line 333: There is one weakness of this approach. If `sampling_interval` was configured in the initial agent configuration, then we have no way to fall back to it, unless we specify the same value in next remote config.
But I'm not sure if it really is an issue.

robsunday (approved):
Besides one doubt - it looks ok

specification/opamp_datamodel.md
ysolomchenko: I'll address it in a separate PR.

specification/opamp_datamodel.md
```diff
@@ -315,20 +315,41 @@ distribution:
         cpu_profiler:
           sampling_interval: 1001
         memory_profiler:
+      callgraphs:
+        sampling_interval: 10
```
pellared, line 319: what is the unit?

ysolomchenko: `ms`, the same as in the file-based configuration:
https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#file-based-configuration

and in the environment variable configuration:
https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#snapshot-profiler

pellared: I think it is better to define it explicitly.

ysolomchenko: Added in [ef36032](https://github.com/signalfx/gdi-specification/pull/406/commits/ef360323fd4b156684178b03a8cfc8edebe3f712)

specification/opamp_datamodel.md
```diff
@@ -315,20 +315,41 @@ distribution:
...
+* When `callgraphs` is present, it indicates that the snapshot profiler SHOULD
+  be started if it is not currently running.
+* When `callgraphs` is omitted, it indicates that the snapshot profiler SHOULD
+  be stopped if it is currently running.
+* When `callgraphs.sampling_interval` is present, it specifies the snapshot
+  profiler's call stack sampling interval in milliseconds.
+* When `callgraphs.sampling_interval` is omitted but `callgraphs` is present,
+  it indicates that the snapshot profiler is enabled with the default sampling
+  interval.
+* When `callgraphs.selection_probability` is present, it indicates the
```
MrAlias, line 347: `callgraphs.selection_probability` is now remotely configurable, but neither effective-configuration field list includes it. That leaves tension between the requirement to reflect remote changes and the guidance not to report additional environment fields.

Should we add `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY` and `distribution.splunk.profiling.callgraphs.selection_probability` to the corresponding lists and examples so the wire representation is explicit?

ysolomchenko: `SPLUNK_SNAPSHOT_SELECTION_PROBABILITY` is expected to be part of Remote Config, I think it would be better to update the effective config to include it instead of removing it from Remote Config. 

I will do it in separate PR.

MrAlias: Agreed—the selection probability should remain in remote config. I wasn’t suggesting removing it; my point was that effective config also needs an explicit representation so the requirement to reflect remote changes is unambiguous. A separate PR works for me. Could you open it and link it here before this merges?

ysolomchenko: https://github.com/signalfx/gdi-specification/pull/409

MrAlias (approved):

breedx-splk (approved):

PR #408, opened 2026-08-06
Introducing specs for Combined Agent

PeterF778:
Initial version of the description.


specification/combined_agent.md
```diff
@@ -0,0 +1,63 @@
...
+| `appd` | The agent uses AppD specific instrumentation only and sends data to an AppD controller. The OTel specific instrumentation is not applied. OTel signals are not generated. | This is the default mode if environment variable `APPDYNAMICS_OPENTELEMETRY_ENABLED` (or system property `appdynamics.opentelemetry.enabled` for Java) is `false` or is not specified. |
+| `dual` | The agent works in **Dual Signal** mode: it uses AppD specific instrumentation to collect APM data for an AppD controller, and uses OTel specific instrumentation to generate OTel signals and sends them to a collector. | This is the default mode if environment variable `APPDYNAMICS_OPENTELEMETRY_ENABLED` (or system property `appdynamics.opentelemetry.enabled` for Java) is `true`. |
+| `hybrid` | The agent works in **Hybrd** mode, using AppD specific instrumentation only. It sends APM data to an AppD controller and generates OTel signals from the same instrumentation sending them to an OTel collector. The OTel specific instrumentation is not applied. | The OTel spans and span attributes can be substantially different from those generated by OTel native instrumentation. |
```
Kielek, line 28: Could you please explain why we need this mode on the instrumentation side?
IMO it better fits to the AppD controller configuration. How you want to propagate this information to the AppD controller to made the magic and translate AppD --> OTel?

PeterF778: The Hybrid mode is another legacy mode that we still support. It is listed here to emphasize the difference between Hybrid and Dual Signal mode, as these two got sometimes misunderstood and mixed up.
I do not understand your last question - nothing on the controller needs to know about Hybrid mode.

MrAlias (commented):
I’m confirming that the proposed requirements align with the existing Combined Agent product and architecture decisions before completing my review. If there is public design context appropriate to reference from this specification, could you point me to it? I’ll follow up offline for any internal context.

specification/combined_agent.md
Kielek: I have need to read it once more time, my first impression was that the AppD controller is responsible for translation AppD Instrumentation data to the OTel one. Sorry for this.

PeterF778:
> If there is public design context appropriate to reference from this specification, could you point me to it? I’ll follow up offline for any internal context.

The only public documentation related to this topic is here: https://help.splunk.com/en/appdynamics-saas/application-performance-monitoring/26.6.0/splunk-appdynamics-for-opentelemetry
Some details are also provided on the child pages.

breedx-splk (approved):

breedx-splk:
@signalfx/gdi-specification-maintainers @signalfx/gdi-specification-approvers can we get another look please? Thanks.

Kielek:
@iskiselev, could you please check? Changes looks fine to me, but you are much more involved in Combined Agent implementation on .NET side.

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+signal processing and exporting as specified by the corresponding Open Telementry
+[documentation for SDK](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)
+and also environment variables that apply to the Open Telemetry autoinstumenting
+agent for a given platform.
+
+## Automatic Agent Configuration Migration
+
+In Dual Signal mode or in OTel-only mode, when the user does not provide their own
+_service namespace_, _service name_ or _deployment environment name_
+for the OTel Resource, the Combined Agent SHOULD configure them automatically
```
iskiselev, line 51: Let's add something like "when possible".
It is not always achievable - for .NET agent some configurations may be discovered post initialization, where we will not be able to modify already discovered resources.

Kielek: _SHOULD_ do the trick here.

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+  },
+  "MD033": {
+    "table_allowed_elements": [ "br" ]
+  }
+}
+-->
+
+# Combined Agent
+
+**Status**: [Experimental](../README.md#versioning-and-status-of-the-specification)
```
MrAlias, line 14: The new document is not added to README's list of specification sections in scope, and this normative change is absent from `CHANGELOG.md`. Could we add both entries so `combined_agent.md` is included in the published scope and version history?

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+
+Combined Agents are targeting AppDynamics customers who wish to migrate their APM
+solution to Splunk O11y Cloud.
+These agents offer the traditional AppDynamics functionality but can be also configured
+to generate Open Telemetry signals.
+
+## Combined Agent Deployment Mode
+
+The functionality of a Combined Agent is defined by its _deployment mode_.
+The deployment mode is specified by environment variable `AGENT_DEPLOYMENT_MODE`.
```
MrAlias, line 24: `AGENT_DEPLOYMENT_MODE` has four table values, but no behavior is defined for an empty or unrecognized value. Since the later precedence rule also says the legacy selector is ignored whenever this variable is "used," `AGENT_DEPLOYMENT_MODE=duel` can leave implementations choosing different modes—or no mode at all. Could we define parsing and fallback/error behavior, including whether values are case-sensitive? The linked OpenTelemetry enum guidance uses a warning plus graceful ignore for unknown values if we want to follow that precedent.

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+| `hybrid` | The agent works in **Hybrd** mode, using AppD specific instrumentation only. It sends APM data to an AppD controller and generates OTel signals from the same instrumentation sending them to an OTel collector. The OTel specific instrumentation is not applied. | The OTel spans and span attributes can be substantially different from those generated by OTel native instrumentation. |
+| `otel` | The agent works in **OTel-only** mode: it uses OTel specific instrumentation, generates OTel signals and emits them to a collector. The agent does not apply AppD specific instrumentation and does not attempt to register with any controller. | |
+
+If the environment variable `AGENT_DEPLOYMENT_MODE` (or system property
+`agent.deployment.mode`) is used, the environment variable
+`APPDYNAMICS_OPENTELEMETRY_ENABLED` and system property
+`appdynamics.opentelementry.enabled` are ignored.
```
MrAlias: This spelling names a different Java system property than the one used in the mode table (`appdynamics.opentelemetry.enabled`). An implementation following this precedence rule would leave the real legacy property active when `agent.deployment.mode` is set.

```suggestion
`appdynamics.opentelemetry.enabled` are ignored.
```

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+
+In Dual Signal or OTel-only modes, the agent MUST accept and handle all
+environment variables that control the OTel specific instrumentation,
+signal processing and exporting as specified by the corresponding Open Telementry
+[documentation for SDK](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)
+and also environment variables that apply to the Open Telemetry autoinstumenting
+agent for a given platform.
```
MrAlias: This `MUST` makes every Combined Agent implement the entire, evolving SDK environment-variable catalog, including signal processors and exporters the platform's embedded SDK may not provide. The linked upstream spec explicitly makes env-var configuration optional and conditions some variables on implemented capabilities. Please scope this to variables applicable to the corresponding Splunk OTel distribution, while requiring upstream names and parsing for the variables it supports.

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+If the environment variable `AGENT_DEPLOYMENT_MODE` (or system property
+`agent.deployment.mode`) is used, the environment variable
+`APPDYNAMICS_OPENTELEMETRY_ENABLED` and system property
+`appdynamics.opentelementry.enabled` are ignored.
+
+In Dual Signal or OTel-only modes, the agent MUST accept and handle all
+environment variables that control the OTel specific instrumentation,
+signal processing and exporting as specified by the corresponding Open Telementry
+[documentation for SDK](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)
+and also environment variables that apply to the Open Telemetry autoinstumenting
```
MrAlias: Use the standard spelling and OpenTelemetry terminology.

```suggestion
and also environment variables that apply to the OpenTelemetry auto-instrumentation
```

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+agent for a given platform.
+
+## Automatic Agent Configuration Migration
+
+In Dual Signal mode or in OTel-only mode, when the user does not provide their own
+_service namespace_, _service name_ or _deployment environment name_
```
MrAlias: These refer to exact OpenTelemetry resource attributes, so code formatting and their canonical names would make the configuration contract unambiguous.

```suggestion
`service.namespace`, `service.name`, or `deployment.environment.name`
```

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+solution to Splunk O11y Cloud.
+These agents offer the traditional AppDynamics functionality but can be also configured
+to generate Open Telemetry signals.
+
+## Combined Agent Deployment Mode
+
+The functionality of a Combined Agent is defined by its _deployment mode_.
+The deployment mode is specified by environment variable `AGENT_DEPLOYMENT_MODE`.
+For Combined Java Agent, the deployment mode can also be specified by system property
+`agent.deployment.mode`, which takes precedence over the environment variable.
```
MrAlias: This document describes Combined Agents as a cross-language concept, but the common configuration contract incorporates Java system properties without identifying the supported implementations or defining their platform-specific capabilities. Please separate the portable deployment-mode requirements from language-specific configuration and state which agents this specification applies to. Runtimes with different mechanisms, such as .NET, can then define their own profile, while runtimes without a Combined Agent can be explicitly out of scope.

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+if it crosses AppD Application boundary.
+* Differences between AppDynamics and OTel instrumentation can cause a single BT
+instance to be composed of multiple traces.
+
+## Contributing to customer telemetry
+
+When a Combined Agent starts in Dual Signal mode, it SHOULD add Open Telemetry
+Resource attribute `appdynamics.agent.version` and set it to the actual
+Combined Agent version.
+For example, for Combined Java Agent, such value might be `25.3.0.36916`.
```
MrAlias, line 88: `appdynamics.agent.version` identifies only a version; it does not identify which AppDynamics agent produced it, and "Combined Agent version" does not say whether this is the AppDynamics component, the embedded Splunk OTel distribution, or the combined package. Please define that value precisely and specify how the agent implementation is identified. The existing `telemetry.sdk.language`, `telemetry.distro.name`, and `telemetry.distro.version` attributes may already provide the portable identity model.

specification/combined_agent.md
```diff
@@ -0,0 +1,88 @@
...
+AppD snapshots captured in Dual Signal mode SHOULD provide the corresponding current
+`TraceId` under the `Data Collectors` section.
+
+Notes:
+
+* This functionality assumes certain alignment of AppDynamics and OTel
+instrumentation and is provided as the _best effort_ only.
+* A single OTel trace can contain multiple BTs,
+if it crosses AppD Application boundary.
+* Differences between AppDynamics and OTel instrumentation can cause a single BT
```
MrAlias: `BT instance` is not self-explanatory to readers unfamiliar with AppDynamics terminology. Spell out the term here.

```suggestion
* Differences between AppDynamics and OTel instrumentation can cause a single Business Transaction
```

specification/combined_agent.md
PeterF778: Working on it.

specification/combined_agent.md
PeterF778: Good catch, thanks!

specification/combined_agent.md
PeterF778: I looked for a comprehensive list of environment variables accepted and supported by Splunk OTel Java Agent, but could not find it (not that I'm really surprised). Perhaps we can remove this paragraph, as it actually does not entail any additional requirements on the Combined Agent.

specification/combined_agent.md
PeterF778: Yes, this is a true gap in the specification, I'm addressing it.

PR #410, opened 2026-08-17
[RemoteConfig] Add empty `config_map` behavior

ysolomchenko:
## Why

https://github.com/signalfx/gdi-specification/pull/406#discussion_r3691211815

Currently, we don't have the possibility to fall back to the startup values, so I think one option could be to treat an empty config map as a signal to use the initial agent configuration.

### DIFF
--- a/specification/opamp_datamodel.md
+++ b/specification/opamp_datamodel.md
@@ -306,6 +306,14 @@ remote configuration. After that, the accepted remote configuration is the
 authoritative source for whether profilers are running and for the profiler
 settings it contains.
 
+An `AgentRemoteConfig` is completely empty when its `config.config_map`
+contains no entries. When an agent receives a completely empty remote
+configuration, it MUST discard any previously accepted remote configuration
+and restore all profiling settings represented by this schema to the values
+from the initial agent configuration established at startup. The initial agent
+configuration remains authoritative until the agent accepts another non-empty
+remote configuration.
+
 ### Data Format
 
 When agents receive a remote configuration with the key `splunk.remote.config` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,8 @@
 - Add support for remote config.
   [#396](https://github.com/signalfx/gdi-specification/pull/396),
   [#402](https://github.com/signalfx/gdi-specification/pull/402),
-  [#406](https://github.com/signalfx/gdi-specification/pull/406)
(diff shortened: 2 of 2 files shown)


specification/opamp_datamodel.md
```diff
@@ -306,6 +306,14 @@ remote configuration. After that, the accepted remote configuration is the
 authoritative source for whether profilers are running and for the profiler
 settings it contains.
 
+An `AgentRemoteConfig` is completely empty when its `config.config_map`
+contains no entries. When an agent receives a completely empty remote
+configuration, it MUST discard any previously accepted remote configuration
+and restore all profiling settings represented by this schema to the values
+from the initial agent configuration established at startup. The initial agent
+configuration remains authoritative until the agent accepts another non-empty
+remote configuration.
```
MrAlias, line 315: `AgentConfigMap` is a collection of independently named configuration objects, and this document permits keys other than `splunk.remote.config`. If a server removes `splunk.remote.config` while retaining another object, the map remains non-empty and this reset never runs, so implementations can keep stale profiling overrides after the current remote configuration no longer contains them. Could we key the fallback to the absence of `splunk.remote.config` instead?

```suggestion
When an `AgentRemoteConfig.config.config_map` contains no
`splunk.remote.config` entry, the agent MUST discard any previously accepted
`splunk.remote.config` and restore all profiling settings represented by this
schema to the values from the initial agent configuration established at
startup. The initial agent configuration remains authoritative until the agent
accepts a remote configuration containing a `splunk.remote.config` entry.
```

MrAlias (commented):
The reset condition does not cover every valid `AgentConfigMap` that omits `splunk.remote.config`, so implementations can diverge when other OpAMP configuration objects are present.

specification/opamp_datamodel.md
```diff
@@ -306,6 +306,13 @@ remote configuration. After that, the accepted remote configuration is the
 authoritative source for whether profilers are running and for the profiler
 settings it contains.
 
+When an `AgentRemoteConfig.config.config_map` contains no
+`splunk.remote.config` entry, the agent MUST discard any previously accepted
+`splunk.remote.config` and restore all profiling settings represented by this
```
robsunday:
```suggestion
`splunk.remote.config` and restore all settings represented by this
```

MrAlias (approved):

robsunday (approved):

